PR #4, opened 2016-11-25
draft bigdata

Eyasics:
propose a skeleton, add MapReduce and part of FlumeJava


heathermiller:
Hi @Eyasics and @msabhi, this PR still contains @cnnrznn's commits (which don't build). Could you please make a new branch and either cherry-pick or make a new commit containing all of yours (but not @cnnrznn's) changes?

msabhi:
Sure Professor, working on it.

msabhi:
Hello Professor @heathermiller ,

I have now excluded @cnnrznn's changes and committed to our branch Jingjing-Abhilash-bigdata. I am sorry for the inconvenience caused. Please let me know if its still causing any problems.

Thanks,
Abhilash

chapter/8/big-data.md
```diff
@@ -1,11 +1,229 @@
...
 ---
+## Introduction
+`JJ: Placeholder for introduction` The booming Internet has generated big data...
 
-Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum. {% cite Uniqueness --file big-data %}
 
-## References
+This chapter is organized in <label for="note1" class="margin-toggle sidenote-number"></label><input type="checkbox" id="note1" class="margin-toggle"/><span class="sidenote">JJ: need to fill in more stuff</span>
+
+- **Data paralleling**:
```
heathermiller: By "Data Paralleling" do you mean the a programming model for data parallelism?

chapter/8/big-data.md
```diff
@@ -1,11 +1,229 @@
...
+  - MapReduce {% cite dean2008mapreduce  --file big-data %}
+  - FlumeJava {% cite chambers2010flumejava --file big-data %}
+  - ...
+- **Graph paralleling**:
```
heathermiller: Likewise, a program model for large-scale parallelism on graphs? (Rather than "Graph paralleling" which is grammatically incorrect)

chapter/8/big-data.md
```diff
@@ -1,11 +1,229 @@
...
+
+## Map Reduce inspired large scale data processing systems
+
+**Dryad/DryadLinq** : 
+
+**Spark (big one)** : 
+
+## Declarative interfaces for the Map Reduce framework
+Map reduce provides only two high level primitives - map and reduce; that the programmers have to worry about. Map reduce takes care of all the processing over a cluster, failure and recovery, data partitioning etc. However, the framework still suffers from rigidity with respect to its one-input data format (key/value pair) and two-stage data flow. Several important patterns like joins (which could be highly complex depending on the data) are extremely hard to implement and reason about for a programmer. Sometimes the code could be become repetitive  when the programmer wants to implement most common operations like projection, filtering etc.
+Non-programmers like data scientists would highly prefer SQL like interface over a cumbersome and rigid framework. Such a high level declarative language can easily express their task while leaving all of the execution optimization details to the backend engine. Also, these kind of abstractions provide ample opportunities for query optimizations.
```
heathermiller: > Non-programmers like data scientists would highly prefer SQL like interface over a cumbersome and rigid framework.

You should not make claims like these without citations or stronger argumentation. It's too hand-wavy as-is.

chapter/8/big-data.md
```diff
@@ -1,11 +1,229 @@
...
+- `combineValues()`, semantically a special case of `parallelDo()`, a combination of a MapReduce combiner and a MapReduce reducer, which is more efficient than doing all the combining in the reducer.
+
+***Deferred Evaluation*** 
+`(JJ: placehoder) join, deferred/materialized; execution plan; figure 1 initial execution plan`
+
+***Optimizer***  
+`(JJ: placehoder) parallelDo Fusion; MSCR;  overall goal to produce the fewest, most efficient MSCR operations in the final optimized plan`
+
+
+**Pig Latin** : Pig latin: a not-so-foreign language for data processing. In SIGMOD, pages 1099–1110, 2008.
```
heathermiller: This is a query language API over map reduce. It's quite different than the others listed here. FlumeJava is a data parallel programming model, whereas Pig Latin is more of like a query language (think SQL)

chapter/8/big-data.md
```diff
@@ -1,11 +1,229 @@
...
 
-{% bibliography --file big-data %}
\ No newline at end of file
+**Hive** :
```
heathermiller: Same point as above. This is out of place here.

heathermiller (commented):
@msabhi and @Eyasics, you guys really need to work more on organization. Right now, as I see it, there is very little structure to this. Topics that are very different are all lumped under the same category, and I don't see a central point arising here. 

This really needs a lot more work! It's not even quite an acceptable rough draft, guys. This really isn't acceptable, especially given that this is the work of _two_ people, not just one. 

You both should come by my office on Monday morning to try and work together to sort this out.

chapter/8/big-data.md
```diff
@@ -1,11 +1,229 @@
...
+
+- **Data paralleling**:
+  - MapReduce {% cite dean2008mapreduce  --file big-data %}
+  - FlumeJava {% cite chambers2010flumejava --file big-data %}
+  - ...
+- **Graph paralleling**:
+  - Pregel
+  - ...
+
+For each programming model, we will discuss the motivation, basic model, execution model, fault-tolerance and performance.
```
heathermiller: Why would you introduce each programming model one by one? It's better to make categories, like "data parallelism" and then to put systems with data parallel programming models into that category.

Also, there are (1) programming models and (2) underlying models of execution. These are hugely important dimensions that need to be treated separately. So far you've lumped everything all together and you're just making a list devoting a little paragraph to each model and not really making this distinction at all. 

Instead you should probably have subsections devoted to programming models and execution models (e.g., spark executes in-memory while Hadoop/MR writes intermediate data to disk)

heathermiller: ...Note that they execute quite differently, but their programming models are quite similar. Do you see my point?

Eyasics: I agree on the comments, and we are going to the programming vs. execution direction. We had a discussion today about this. The push was from a while ago and I didn't really have time to update that, sorry for the confusion, and I am working on writing introduction (discussions over models to cover and relationships between them) now, and will address these points.

chapter/8/big-data.md
msabhi: I am sorry professor, the section "Introduction" shouldn't have been there. It missed my eyes. Its speaking about the old draft and as we can see the new draft could be like below.
 - Map Reduce programming model
 - Problems with Iterative map reduce and new frameworks to handle such iterative algorithms
 - Spark & DyradLinq - A new functional programming way of doing Map reduce
 - Declarative interfaces - easy programming and opportunities for optimizations
- A special model - SparkSQL which enables the intermixing of both relational and procedural
 - Frameworks with optimizers
 - A brief introduction to a parallel world of large scale graph processing -> their limitations -> and how GraphX overcame it by leveraging Spark to address graph processing algorithms.

I am really sorry for this. I am working on this with Jingjing. We will meet you this monday to revisit the flow.

chapter/8/big-data.md
msabhi: Sure professor. I will include suggestions from papers to strengthen this argument.

chapter/8/big-data.md
msabhi: In this section, I am talking about SQL-like interfaces. So, yes professor, we must have FlumeJava in data-parallel pipeline. I will make those changes.

chapter/8/big-data.md
msabhi: It should have been HiveQL instead of just Hive.

chapter/8/big-data.md
heathermiller: Ok, this is getting better, but it still seems a bit confused. Maybe you've understood this now already from my other comments, by why not break things up differently? 

* Programming Models
  - Data parallelism (most popular, standard map/reduce/functional pipelining)
     - Limitations, iteration difficult due to the execution model of MapReduce/Hadoop
  - Graphs
  - Querying

* Execution Models
  - MapReduce (intermediate writes to disk)
    - Limitations, iteration, performance
  - Spark (all in memory)
    - Limitations ?

* Things people are building on top of MapReduce/Spark
  - FlumeJava? ...Etc
  - Ecosystem, everything interoperates with GFS or HDFS, or makes use of stuff like protocol buffers so systems like Pregel and MapReduce and even MillWheel...

chapter/8/big-data.md
heathermiller: Sure but my point is that these things are very very different. They shouldn't be logically lumped together. You should reorganize your structure (see above)

chapter/8/big-data.md
heathermiller: Sure, but it's still a query language on top of Hadoop next to FlumeJava which is very different. I'm just saying these should be in separate sections.

chapter/8/big-data.md
Eyasics: I am also thinking of adding one more discussion about scheduler and optimizer (call it performance section)

chapter/8/big-data.md
msabhi: I agree professor. We will make changes.

chapter/8/big-data.md
heathermiller: Ah yeah, of course. That would be a great idea. Definitely (sorry I omitted it)

chapter/8/big-data.md
```diff
@@ -1,11 +1,332 @@
...
+    - Dryad: What if we think individual computation tasks as vertices? We essentially construct a communication graph between those vertices. What programmers need to do is to describe this DAG graph and let Dryad execution engine to construct the execution plan and take care of scheduling. Like MP, writing raw Dryad is hard, programmers need to understand system resources and other lower-level details. This motivates a more declarative programming model: DryadLINQ as a querying language.   
+    `Q: Should this go to execution model?`
+    - Spark: what is Spark? how is it different from map reduce? (RDD/lineage: can support iterative algorithm, interactive analytics;) what is pipelining? why is Spark so powerful - RDD and API? What is a RDD and why is it so efficient? properties of a RDD? why is RDD better than DSM? What are the transformations and actions available in Spark ?
```
heathermiller: Can't FlumeJava also do the same sort of pipelining?

chapter/8/big-data.md
```diff
@@ -1,11 +1,332 @@
...
+  - Large-scale Parallelism on Graphs
+    - Why a separate graph processing model? what is a BSP? working of BSP? Do not stress more since its not a map reduce world exactly.
+    - GraphX programming model (working on this)
+  - Querying: more declarative
+    - DryadLINQ: SQL-like, uses Dryad as execution engine;   
+    `Suggestion: Merge this with Dryad above?`
+    - Pig: on top of Hadoop, independent of execution platform, in theory can compiled into DryadLINQ too; what is the performance gain/lost? Easier to debug?
```
heathermiller: Isn't Sawzall also query-esque?

chapter/8/big-data.md
```diff
@@ -1,11 +1,332 @@
...
+    `Q: Should this go to execution model?`
+    - Spark: what is Spark? how is it different from map reduce? (RDD/lineage: can support iterative algorithm, interactive analytics;) what is pipelining? why is Spark so powerful - RDD and API? What is a RDD and why is it so efficient? properties of a RDD? why is RDD better than DSM? What are the transformations and actions available in Spark ?
+  - Large-scale Parallelism on Graphs
+    - Why a separate graph processing model? what is a BSP? working of BSP? Do not stress more since its not a map reduce world exactly.
+    - GraphX programming model (working on this)
+  - Querying: more declarative
+    - DryadLINQ: SQL-like, uses Dryad as execution engine;   
+    `Suggestion: Merge this with Dryad above?`
+    - Pig: on top of Hadoop, independent of execution platform, in theory can compiled into DryadLINQ too; what is the performance gain/lost? Easier to debug?   
+    `Q: Do we need to include Hive?`
```
heathermiller: Yes. Hive is a hugely important project.

chapter/8/big-data.md
```diff
@@ -1,11 +1,332 @@
...
+
+- Execution Models
+  - MapReduce (intermediate writes to disk): What is the sequence of actions when a MapReduce functions are called? How is write-to-disk good/bad (fault-tolerant/slow)? How does the data are transmitted across clusters efficiently (store locally)? To shorten the total time for MP operations, it uses backup tasks. When MP jobs are pipelined, what optimizations can be performed by FlumeJava? In spite of optimizations and pipelining, what is the inherent limitation (not support iterative algorithm?)
+  - Spark (all in memory): introduce spark architecture, different layers, what happens when a spark job is executed? what is the role of a driver/master/worker, how does a scheduler schedule the tasks and what performance measures are considered while scheduling? how does a scheduler manage node failures and missing partitions? how are the user defined transformations passed to the workers? how are the RDDs stored and memory management measures on workers? do we need checkpointing at all given RDDs leverage lineage for recovery? if so why ?
+  - Graphs :
```
heathermiller: Shouldn't this be in a section on programming models? Sure, Pregel is both programming model + model of execution that's different from everything else. But Spark has shown that it's possible to run graph processing algorithms on top of Spark faster than other frameworks, no? And it interoperates nicely with Spark and all of the other APIs that sit on top of Spark. So it seems to me that Pregel might be a little less important these days than it once was, so perhaps it's a distraction to spend a lot of time on Pregel's model of execution since every other framework compiles to Spark or MR.

chapter/8/big-data.md
```diff
@@ -1,11 +1,332 @@
...
+  - Introducing streaming: what happens when data cannot be complete? How does different programming model adapt? windowing `todo: more`
+
+## Programming Models
+### Data parallelism
+The motivation for MapReduce {% cite dean2008mapreduce  --file big-data %} is that we want to use hundreds/thousands of machines to do data processing in parallel, but we don’t want to deal with low-level management. MapReduce can help this by abstracting computing logic into simple map and reduce functions and let the computation model handle the parallelization and distribution, provide fault tolerance, manage I/O scheduling and get proper status updates. The solution in the MapReduce paper is simple and powerful in terms of separating programming model and the executing model. This model applies to computations that are usually parallelizable: A `map` function can operate on each logical "record", this generates a set of intermediate key/value pairs, and then a `reduce` function applies on all values that share the same key and generate one or zero output value. Conceptually, the map and reduction functions have associated **types**:
```
heathermiller: > but we don’t want to deal with low-level management.
Low-level management of what?

heathermiller: > A `map` function can operate on each logical "record"

What, concretely is a record?

heathermiller: > then a `reduce` function applies on all values that share the same key and generate one or zero output value

Grammar. Sentence needs work, grammatically and terminologically incorrect.

chapter/8/big-data.md
```diff
@@ -1,11 +1,332 @@
...
+```
+map (k1,v1) -> → list(k2,v2)
+reduce (k2,list(v2)) -> list(v2)
```
heathermiller: To a PL person, these don't look like types. Is this copied/pasted from the paper? If so, I recall that it was in MathTeX, right? Just to not irritate people, I would use LaTeX here rather than code.

chapter/8/big-data.md
```diff
@@ -1,11 +1,332 @@
...
+```
+The input keys and values are drawn from a different domain than the output keys and values. The intermediate keys and values are from the same domain as the output keys and values.
```
heathermiller: This is not the data parallelism programming model. It's the MapReduce programming model. Perhaps read a bit about data parallelism.

heathermiller: Also, you should include other data parallel models here. Not just MapReduce. Spark is data parallel.

chapter/8/big-data.md
```diff
@@ -1,11 +1,332 @@
...
+
+**Execution** At high level, when the user program calls *MapReduce* function, the input files are split into *M* pieces and it runs *map* function on corresponding splits; then intermediate key space are partitioned into *R* pieces using a partitioning function; After the reduce functions all successfully complete, the output is available in *R* files. The sequences of actions are shown in the figure below. We can see from label (4) and (5) that the intermediate key/value pairs are written/read into disks, this is a key to fault-tolerance in MapReduce model and also a bottleneck for more complex computation algorithms.
```
heathermiller: You don't explain fault tolerance at all. You really should.

chapter/8/big-data.md
```diff
@@ -1,11 +1,332 @@
...
+
+<figure class="main-container">
+  <img src="{{ site.baseurl }}/resources/img/mapreduce-execution.png" alt="MapReduce Execution Overview" />
+</figure>
+
+**Limitations & Extensions**  
+***Real-world applications often require a pipeline of MapReduce jobs and the management becomes an issue.***  
+**FlumeJava** was introduced to make it easy to develop, test, and run efficient data-parallel pipelines. FlumeJava represents each dataset as an object and transformation is invoked by using methods on these objects. It constructs an efficient internal execution plan from a pipeline of MapReduce jobs using deferred evaluation and optimizers such as fusions. The debugging ability allows programmers to run on the local machine first and then deploy to large clusters.
```
heathermiller: Why is this introduced here? Introducing FlumeJava here doesn't make a lot of sense. This should be the section on Limitations and Extensions, right? Where's the limitation?

heathermiller: "the management becomes an issue" is not clear enough of an explanation.

chapter/8/big-data.md
```diff
@@ -1,11 +1,332 @@
...
+
+**Dryad/DrydaLINQ** Dryad is a more general and flexible execution engine that execute subroutines at a specified graph vertices. Developers can specify an arbitrary directed acyclic graph to combine computational "vertices" with communication channels (file, TCP pipe, shared-memory FIFO) and  build a dataflow graph. Compared with MapReduce, Dryad can specify an arbitrary DAG that have multiple number of inputs/outputs and support multiple stages. Also it can have more channels and boost the performance when using TCP pipes and shared-memory. But like writing a pipeline of MapReduce jobs, Dryad is a low-level programming model and hard for users to program, thus a more declarative model - DryadLINQ was created to fill in the gap. It exploits LINQ, a query language in .NET and automatically translates the data-parallel part into execution plan and passed to the Dryad execution engine.
```
heathermiller: Why is this here? It's unrelated to everything before and after it.

chapter/8/big-data.md
```diff
@@ -1,11 +1,332 @@
...
+This model was introduced in 1980 to represent the hardware design features of parallel computers. It gained popularity as an alternative for map reduce since it addressed the above mentioned issues with map reduce<br />
+BSP model is a message passing synchronous model where -
+
+ - Computation consists of several steps called as supersets.
+ - The processors involved have their own local memory and every processor is connected to other via a point-to-point communication.
+ - At every superstep, a processor receives input at the beginning, performs computation and outputs at the end.
+ - A processor at superstep S can send message to another processor at superstep S+1 and can as well receive message from superstep S-1.
+ - Barrier synchronization synchs all the processors at the end of every superstep.
+
+A notable feature of the model is the complete control on data through communication between every processor at every superstep. Though similar to map reduce model, BSP preserves data in memory across supersteps and helps in reasoning iterative graph algorithms.
```
heathermiller: BSP is a model of execution, not a programming model.

chapter/8/big-data.md
```diff
@@ -1,11 +1,332 @@
...
+
+### Querying
```
heathermiller: Is this supposed to be about query-languages as a programming models on top of MR?

chapter/8/big-data.md
```diff
@@ -1,11 +1,332 @@
...
+
+
+
+## Execution Models
```
heathermiller: Good, though this section's existence seems out of sync with sections above.

chapter/8/big-data.md
```diff
@@ -1,11 +1,332 @@
...
+
+
+- Actions - operations that trigger computation on a RDD and return values.
+
+  - `reduce(f:(T,T)⇒T) : RDD[T] ⇒ T` : return T by reducing the elements using specified commutative and associative binary operator
+  - `collect()` : Return an Array[T] containing all elements
+  - `count()` : Return the number of elements
+
+
+Why RDD over Distributed Shared memory (DSM) ?
```
heathermiller: Good question! I'm happy you realized to bring up this argument.

chapter/8/big-data.md
```diff
@@ -1,11 +1,332 @@
...
+RDDs are immutable and can only be created through coarse grained transformation while DSM allows fine grained read and write operations to each memory location. Hence RDDs do not incur the overhead of checkpointing thats present in DSM and can be recovered using their lineages.
+RDDs are immutable and hence a straggler(slow node) can be replaced with backup copy as in Map reduce. This is hard to implement in DSM as two copies point to the same location and can interfere in each other’s update.
+Other benefits include the scheduling of tasks based on data locality to improve performance and the ability of the RDDs to degrade gracefully incase of memory shortage. Partitions that do not fit in RAM gets spilled to the disk (performance will then be equal to that of any data parallel system).
+
+
+- Pig/HiveQL/SparkSQL
+  - Limitations ?
+
+**Pregel**
```
heathermiller: Is this supposed to be a subsection of the Execution Models stuff?

chapter/8/big-data.md
```diff
@@ -1,11 +1,332 @@
...
+
+
+## Iterative processing in Map Reduce
+
+Many a analytics workloads like K-means, logistic regression, graph processing applications like PageRank, shortest path using parallel breadth first search require multiple stages of map reduce jobs. In regular map reduce framework like Hadoop, this requires the developer to manually handle the iterations in the driver code. At every iteration, the result of each stage T is written to HDFS and loaded back again at stage T+1 causing a performance bottleneck. The reason being wastage of network bandwidth, CPU resources and mainly the disk I/O operations which are inherently slow. In order to address such challenges in iterative workloads on map reduce, frameworks like Haloop, Twister and iMapReduce adopt special techniques like caching the data between iterations and keeping the mapper and reducer alive across the iterations.
+
+
+
+
+**Haloop** : HaLoop: Efficient Iterative Data Processing on Large Clusters.
```
heathermiller: Keep this reference.

chapter/8/big-data.md
```diff
@@ -1,11 +1,332 @@
...
+
+**iMapReduce**: iMapReduce: A Distributed Computing Framework for Iterative Computation
```
heathermiller: Keep this reference.

chapter/8/big-data.md
```diff
@@ -1,11 +1,332 @@
...
+
+**Twister** :  Twister: a runtime for iterative MapReduce.
```
heathermiller: Keep this reference.

chapter/8/big-data.md
```diff
@@ -1,11 +1,332 @@
...
+Non-programmers like data scientists would highly prefer SQL like interface over a cumbersome and rigid framework. Such a high level declarative language can easily express their task while leaving all of the execution optimization details to the backend engine. Also, these kind of abstractions provide ample opportunities for query optimizations.
+
+**Introduce Sazwal** (its now no more used but one of the first ideas) : Parallel analysis with Sawzall. Scientific Programming, 13(4):277–298, 2005
+
+** FlumeJava (2010) **
+
+Many real-world computations involves a pipeline of MapReduces, and this motivates additional management to chain together those separate MapReduce stages in an efficient way. FlumeJava {% cite chambers2010flumejava --file big-data %} can help build those pipelines and keep computations modular. At core, FlumeJava are a couple of classes that represent immutable parallel collections. It defers evaluation and optimization by internally constructing an execution plan dataflow graph.
+
+***Core Abstraction***
```
heathermiller: It might be good to keep this. I would use this when you figure out where to introduce FlumeJava.

chapter/8/big-data.md
```diff
@@ -1,11 +1,473 @@
 ---
 layout: page
 title:  "Large Scale Parallel Data Processing"
-by: "Joe Schmoe and Mary Jane"
+by: "Jingjing and Abhilash"
 ---
+## Introduction
+The growth of Internet has generated the so-called big data(terabytes or petabytes). It is not possible to fit them into a single machine or process them with one single program. Often the computation has to be done fast enough to provide practical services. A common approach taken by tech giants like Google, Yahoo, Facebook is to process big data across clusters of commodity machines. Many of the computations are conceptually straightforward, and Google proposed the MapReduce model to abstract the logic and proved to be simple and powerful. From then on, the idea inspired lots of other programming models. In this chapter, we will present how programming models evolve over time, why their execution engines are designed in certain ways, and underlying ecosystem that supports each developing thread.
```
heathermiller, line 7: When the rest of the chapter is done, I think we need to revisit this introduction paragraph. In particular, I think we should be focusing more on defining exactly what **batch** processing is. As a later step we can mention that large companies need to do this processing, etc.

chapter/8/big-data.md
```diff
@@ -1,11 +1,473 @@
...
+## 1 Programming Models
+### 1.1 Data parallelism
+*Data parallelism* is to run a single operation on different pieces of the data on different machines in parallel. Comparably, a sequential computation looks like *"for all elements in the dataset, do operation A"*, where dataset could be in the order of terabytes or petabytes aka. big data and one wants to scale up the processing. The challenges to do this sequential computation in a parallelized manner include how to abstract the different types of computations in a simple and correct way, how to distribute the data to hundreds/thousands of machines, how to handle failures and so on.
```
heathermiller, line 10: > *Data parallelism* is to run a single operation on different pieces of the data on different machines in parallel.

This definition can be improved. For example, I would maybe define data parallelism as follows:

> *Data parallelism* is, given a dataset, the simultaneous execution on multiple machines or threads of the same function across groups of elements of a dataset. Data parallelism can also be thought of as a subset of SIMD ("single instruction, multiple data") execution, a class of parallel execution in Flynn's taxonomy.

chapter/8/big-data.md
```diff
@@ -1,11 +1,473 @@
...
+
+The MapReduce model is simple and powerful, and quickly became very popular among developers. However, when developers start writing real-world applications, they often end up chaining together MapReduce stages. The pipeline of MapReduce forces programmers to write additional coordinating codes, i.e. the development style goes backward from simple logic computation abstraction to lower-level coordination management. In map reduce, programmers need to reason about data representation on disk or in storage services such as a database. Besides, developers need to clearly understand the map reduce execution model  to do manual optimizations[ref]. **FlumeJava** {%cite chambers2010flumejava --file big-data%} library intends to provide support for developing data-parallel pipelines by abstracting away the complexity involved in data representation and implicitly handling the optimizations. It defers the evaluation, constructs an execution plan from parallel collections, optimizes the plan, and then executes underlying MR primitives. The optimized execution is comparable with hand-optimized pipelines, so there's no need to write raw MR programs directly.
```
heathermiller, line 18: > The pipeline of MapReduce forces programmers to write additional coordinating codes, i.e. the development style goes backward from simple logic computation abstraction to lower-level coordination management.

Improve this explanation of the limitations.

I think a stronger point here is that to write a single MapReduce "job" (of which you need many in a pipeline) you have to write a lot of boilerplate. So you have a bunch of files and a bunch of boilerplate that you have to write for each job in the pipeline. This quickly gets cumbersome.

heathermiller: > In map reduce, programmers need to reason about data representation on disk or in storage services such as a database.

I think it's less about reasoning and more about performance here, right? I mean, sure it's annoying to have to write out intermediate data, but the bigger issue is that having to write all of this data to disk, is painfully slow, though this is of course good for fault tolerance (since they don't do something like lineages). I think you should try to make more points like this here if you're going to list the limitations.

heathermiller: > Besides, developers need to clearly understand the map reduce execution model  to do manual optimizations[ref]

I understand what you're trying to say here, but it's not clear what you mean to someone who doesn't know MapReduce yet. You're going to have to give an example or at least devote another sentence or two to explaining why optimizations are necessary. Only someone who already understands MapReduce would understand the point you're making here.

chapter/8/big-data.md
```diff
@@ -1,11 +1,473 @@
...
+
+An alternative approach to data prallelism is to construct complex, multi-step directed acyclic graphs (DAGs) of work from the user instructions and execute those DAGs all at once. This eliminates the costly synchronization required by MapReduce and makes applications much easier to build and reason about. Dryad, a Microsoft Research project used internally at Microsoft was one such project which leveraged this model of computation.
```
heathermiller, line 20: > An alternative approach to data prallelism is to construct complex, multi-step directed acyclic graphs (DAGs) of work from the user instructions and execute those DAGs all at once.

Make a diagram showing what a DAG is. Again, someone who has never done this kind of programming before won't know what you mean. A picture is an easy way to quickly give them intuition about this.

heathermiller: > This eliminates the costly synchronization required by MapReduce and makes applications much easier to build and reason about.

Is synchronization really the problem? I'm not sure what you mean by this.

chapter/8/big-data.md
```diff
@@ -1,11 +1,473 @@
...
...tional programming and pipelining to provide such support. It is largely inspired by MapReduce's model and builds upon the ideas behind DAG, lazy evaluation of DryadLinq. Instead of writing data to disk for each job as MapReduce does Spark can cache the results across jobs. Spark explicitly caches computational data in memory thorugh specialized immutable datasets named Resilient Distributed Sets(RDD) and reuse the same dataset across multiple parallel operations. The Spark builds upon RDD to achieve fault tolerance by reusing the lineage information of the lost RDD. This results in lesser overhead than what is seen in fault tolerance achieved by checkpoint in Distribtued Shared Memory systems. Moreover, Spark powers a stack of other libraries, e.g..SQL&DataFrames, GraphX, and can easily combine those libraries in one single application. These feature makes Spark the best fit for iterative jobs and interactive analytics and also helps it in providing better performance. Above all, any system can be easily expressed by Spark enabling other models to leverage the specific advantages of Spark systems and still retain the process of computation without any changes to Spark system[ref].
```
heathermiller, line 25: > Moreover, Spark powers a stack of other libraries, e.g..SQL&DataFrames, GraphX, and can easily combine those libraries in one single application.

I would say,

>Moreover, Spark is the underlying framework upon which many very different systems are built, e.g., Spark SQL & DataFrames, GraphX, Streaming Spark, which makes it easy to mix and match the use of these systems all in the same application.

heathermiller: > Above all, any system can be easily expressed by Spark enabling other models to leverage the specific advantages of Spark systems and still retain the process of computation without any changes to Spark system[ref].

Well, this isn't exactly true. I could come up with many systems that cannot be expressed with Spark. I think the point that you want to make is that the MapReduce programming model (representing your computations as map/reduce combinators) turns out to be an incredibly general model, as Spark has demonstrated, by implementing everything from SQL to graphs on top of it.

chapter/8/big-data.md
```diff
@@ -1,11 +1,473 @@
...
+
+### 1.1.2 FlumeJava
+FlumeJava {%cite chambers2010flumejava --file big-data %}was introduced to make it easy to develop, test, and run efficient data-parallel pipelines. FlumeJava represents each dataset as an object and transformation is invoked by applying methods on these objects. It constructs an efficient internal execution plan from a pipeline of MapReduce jobs, uses deferred evaluation and optimizes based on plan structures. The debugging ability allows programmers to run on the local machine first and then deploy to large clusters.
```
heathermiller, line 74: > uses deferred evaluation and optimizes based on plan structures.

It uses deferred evaluation, but only for fault-tolerance, right? I recall that it had nothing to do with staging code the way that Spark does, but I'm forgetting exactly why at present.

chapter/8/big-data.md
```diff
@@ -1,11 +1,473 @@
...
+*Core Abstraction*  
+- `PCollection<T>`, a immutable bag of elements of type `T`
+- `recordOf(...)`, specifies the encoding of the instance
+- `PTable<K, V>`, a subclass of `PCollection<Pair<K,V>>`, a immutable multi-map with keys of type `K` and values of type `V`
+- `parallelDo()`, can be expressed both the map and reduce parts of MapReduce
+- `groupByKey()`, same as shuffle step of MapReduce
+- `combineValues()`, semantically a special case of `parallelDo()`, a combination of a MapReduce combiner and a MapReduce reducer, which is more efficient than doing all the combining in the reducer.
+
+*Deferred Evaluation & Optimizer*  
+The state of each `PCollection` object is either *deferred* (not yet computed) and *materialized* (computed). When the program invokes a parallel operation, it does not actually run the operation. Instead, it performs the operation only when needed. FlumeJava also provides some optimization practices: 1) parallelDo Fusion: f(g(x)) => f o g(x) to reduce steps; 2) MapShuffleCombineReduce (MSCR) Operation that generalizes MapReduce jobs to accept multiple inputs and multiple outputs. And for this, FlumeJava does another MSCR fusion.
```
heathermiller, line 85: > 1) parallelDo Fusion: f(g(x)) => f o g(x) to reduce steps; 2) MapShuffleCombineReduce (MSCR) Operation

Make an ordered list out of this and either use LaTeX syntax or backticks to highlight appropriately.

chapter/8/big-data.md
```diff
@@ -1,11 +1,473 @@
...
+- `PTable<K, V>`, a subclass of `PCollection<Pair<K,V>>`, a immutable multi-map with keys of type `K` and values of type `V`
+- `parallelDo()`, can be expressed both the map and reduce parts of MapReduce
+- `groupByKey()`, same as shuffle step of MapReduce
+- `combineValues()`, semantically a special case of `parallelDo()`, a combination of a MapReduce combiner and a MapReduce reducer, which is more efficient than doing all the combining in the reducer.
+
+*Deferred Evaluation & Optimizer*  
+The state of each `PCollection` object is either *deferred* (not yet computed) and *materialized* (computed). When the program invokes a parallel operation, it does not actually run the operation. Instead, it performs the operation only when needed. FlumeJava also provides some optimization practices: 1) parallelDo Fusion: f(g(x)) => f o g(x) to reduce steps; 2) MapShuffleCombineReduce (MSCR) Operation that generalizes MapReduce jobs to accept multiple inputs and multiple outputs. And for this, FlumeJava does another MSCR fusion.  
+
+
+### 1.1.3 Dryad
```
heathermiller, line 88: This section is actually less detailed than the section above about Dryad. It's also a bit repetitive when comparing to that above Dryad section. Therefore, you should add more detail, give the primitives of the Dryad model and an example or two (same level of depth as the MapReduce section).

chapter/8/big-data.md
```diff
@@ -1,11 +1,473 @@
...
+Dryad is a more general and flexible execution engine that execute subroutines at a specified graph vertices. Developers can specify an arbitrary directed acyclic graph to combine computational "vertices" with communication channels (file, TCP pipe, shared-memory FIFO) and  build a dataflow graph. Compared with MapReduce, Dryad can specify an arbitrary DAG that have multiple number of inputs/outputs and support multiple stages. Also it can have more channels and boost the performance when using TCP pipes and shared-memory. But like writing a pipeline of MapReduce jobs, Dryad is a low-level programming model and hard for users to program, thus a more declarative model - DryadLINQ  {%cite yu2008dryadlinq --file big-data %} was created to fill in the gap. It exploits LINQ, a query language in .NET and automatically translates the data-parallel part into execution plan and passed to the Dryad execution engine. Like MR, writing raw Dryad is hard, programmers need to understand system resources and other lower-level details. This motivates a more declarative programming model: DryadLINQ - a querying language.
```
heathermiller, line 89: > Dryad is a more general and flexible execution engine

More general and flexible than what? Spark? MapReduce? Be explicit about what it's more general/flexible than, and back up why this is the case.

heathermiller: (This is easy to back up, you can just connect with the next sentence by starting the next "This is because...")

heathermiller: > Compared with MapReduce, Dryad can specify an arbitrary DAG that have multiple number of inputs/outputs and support multiple stages.

Assuming the first sentence of this paragraph was referring to Dryad being more flexible than MapReduce, this then becomes repetitive. Eliminate either this sentence or the beginning sentence.

heathermiller: > But like writing a pipeline of MapReduce jobs, Dryad is a low-level programming model and hard for users to program, thus a more declarative model

Be more specific. Why is this the case? This isn't enough detail here.

heathermiller: > It exploits LINQ, a query language in .NET and automatically translates the data-parallel part into execution plan and passed to the Dryad execution engine.

Please show examples of DryadLINQ. You did point out that it's a query language, but you've also shown the primitives of stuff like FlumeJava and MapReduce. DryadLINQ, as a query language, looks totally different than these. Therefore it would be very informative for you to show this to the reader.

heathermiller: > Like MR, writing raw Dryad is hard, programmers need to understand system resources and other lower-level details. This motivates a more declarative programming model: DryadLINQ - a querying language.

This is repeated

chapter/8/big-data.md
```diff
@@ -1,11 +1,473 @@
...
+RDDs are immutable and hence a straggler(slow node) can be replaced with backup copy as in Map reduce. This is hard to implement in DSM as two copies point to the same location and can interfere in each other’s update.
+Other benefits include the scheduling of tasks based on data locality to improve performance and the ability of the RDDs to degrade gracefully incase of memory shortage. Partitions that do not fit in RAM gets spilled to the disk (performance will then be equal to that of any data parallel system).
+
+***Challenges in Spark***
+
+- `Functional API semantics` : The GroupByKey operator is costly in terms of performance. In that it returns a distributed collection of (key, list of value) pairs to a single machine and then an aggregation on individual keys is performed on the same machine resulting in computation overhead. Spark does provide reduceByKey operator which does a partial aggregation on invidual worker nodes before returning the distributed collection. However, developers who are not aware of such a functionality can unintentionally choose groupByKey.
```
heathermiller, line 138: Explain why they might unintentionally choose `groupByKey`! This is a big and important point.

The reason is that when you write a functional program, you think more declaratively about your problem. That is, you think about the sort of _transformation_ you want to do to your data–you think about _what_ you want to do, and not _how_ you want to do it. 

Programmers who have learned a bit of functional programming (common for Scala developers) tend to think only about the end result of the `groupByKey` operator. They've been trained to not have to think how `groupByKey` is implemented atop of the cluster. Therefore, to use Spark, unlike functional programming languages, you need to understand how the underlying cluster is going to execute your code. The burden of saving performance is then left to the programmer, who's expected to understand the underlying execution model of Spark, and who should know when to use `reduceByKey` over `groupByKey`.

chapter/8/big-data.md
```diff
@@ -1,11 +1,473 @@
...
+
+*Distributed in-memory storage - Resilient Distributed Data sets :*
+
+RDD is a partitioned, read only collection of objects which can be created from data in stable storage or by transforming other RDD. It can be distributed across multiple nodes (parallelize) in a cluster and is fault tolerant(Resilient). If a node fails, a RDD can always be recovered using its lineage graph (information on how it was derived from dataset). A RDD is stored in memory (as much as it can fit and rest is spilled to disk) and is immutable - It can only be transformed to a new RDD. These are the lazy transformations which are applied only if any action is performed on the RDD. Hence, RDD need not be materialized at all times.
```
heathermiller, line 97: > RDD can always be recovered using its lineage graph (information on how it was derived from dataset)

I would say,

> an RDD can always be recovered using its lineage; the DAG of computations performed on the source dataset.

heathermiller: > These are the lazy transformations which are applied only if any action is performed on the RDD. Hence, RDD need not be materialized at all times.

I would say,

> These transformations are deferred; that means they are built up and staged, and aren't actually applied until an action is performed on an RDD. Thus, it's important to note that while one might have applied many transformations to a given RDD, some resulting transformed RDD may not be materialized even though one may hold a reference to it.

chapter/8/big-data.md
```diff
@@ -1,11 +1,473 @@
...
+The properties that power RDD with the above mentioned features :
+- A list of dependencies on other RDD’s.
+- An array of partitions that a dataset is divided into.
+- A compute function to do a computation on partitions.
+- Optionally, a Partitioner for key-value RDDs (e.g. to say that the RDD is hash-partitioned)
+- Optional preferred locations (aka locality info), (e.g. block locations for an HDFS file)
+
+
+<figure class="main-container">
+  <img src="./spark_pipeline.png" alt="Spark pipeline" />
```
heathermiller, line 108: Nice 😄

chapter/8/big-data.md
```diff
@@ -1,11 +1,473 @@
...
+- Actions - operations that trigger computation on a RDD and return values.
+
+  - `reduce(f:(T,T)⇒T) : RDD[T] ⇒ T` : return T by reducing the elements using specified commutative and associative binary operator
+  - `collect()` : Return an Array[T] containing all elements
+  - `count()` : Return the number of elements
+
+RDDs by default are discarded after use. However, Spark provides two explicit operations  persist() and cache() to ensure RDDs are persisted in memory once the RDD has been computed for the first time.
+
+*Why RDD over Distributed Shared memory (DSM) ?*
+RDDs are immutable and can only be created through coarse grained transformation while DSM allows fine grained read and write operations to each memory location. Hence RDDs do not incur the overhead of checkpointing thats present in DSM and can be recovered using their lineages.
```
heathermiller, line 132: > Hence RDDs do not incur the overhead of checkpointing thats present in DSM and can be recovered using their lineages.

Well, RDDs don't require checkpointing at all due to the fact that they are immutable and can be derived from lineages. So checkpointing is simply not required at all.

Update the wording here to reflect this.

chapter/8/big-data.md
```diff
@@ -1,11 +1,473 @@
...
+RDDs are immutable and hence a straggler(slow node) can be replaced with backup copy as in Map reduce. This is hard to implement in DSM as two copies point to the same location and can interfere in each other’s update.
+Other benefits include the scheduling of tasks based on data locality to improve performance and the ability of the RDDs to degrade gracefully incase of memory shortage. Partitions that do not fit in RAM gets spilled to the disk (performance will then be equal to that of any data parallel system).
```
heathermiller, line 134: DSM doesn't have some notion of locality that it can also exploit? How does DSM compare here?

chapter/8/big-data.md
```diff
@@ -1,11 +1,473 @@
...
+
+- `Debugging and profiling` : There is no availability of debugging tools and developers find it hard to realize if a computation is happening more on a single machine or if the data-structure they used were inefficient.
+
+### 1.2 Querying: declarative interfaces
+MapReduce provides only two high level primitives - map and reduce that the programmers have to worry about. MapReduce takes care of all the processing over a cluster, failure and recovery, data partitioning etc. However, the framework suffers from rigidity with respect to its one-input data format (key/value pair) and two-stage data flow.
+Several important patterns like joins (which could be highly complex depending on the data) are extremely hard to implement and reason about for a programmer. Sometimes the code could be become repetitive  when the programmer wants to implement most common operations like projection, filtering etc.
```
heathermiller, line 144: > Several important patterns like joins (which could be highly complex depending on the data) are extremely hard to implement and reason about for a programmer.

You mean these are hard to reasonable for a programmer _when that programmer is limited to using only MapReduce_? Please clarify here.

chapter/8/big-data.md
```diff
@@ -1,11 +1,473 @@
...
+Non-programmers like data scientists would highly prefer SQL like interface over a cumbersome and rigid framework{% cite scaling-spark-in-real-world --file big-data%}. Such a high level declarative language can easily express their task while leaving all of the execution optimization details to the backend engine. Hence, these kind of abstractions provide ample opportunities for query optimizations.
+
+Sawzall {% cite pike2005interpreting --file big-data%} is a programming language built on top of MapReduce. It consists of a *filter* phase (map) and an *aggregation* phase (reduce). User program can specify the filter function, and emit the intermediate pairs to external pre-built aggregators.
+
+Apart from Sawzal, Pig  {%cite olston2008pig --file big-data %} and Hive  {%cite thusoo2009hive --file big-data %} are the other major components that sit on top of Hadoop framework for processing large data sets without the users having to write Java based MapReduce code.
+
```
heathermiller, line 150: Can you provide an example of Sawzall, Pig, or Hive code? Just to illustrate how exactly a MapReduce job can be realized in a query language. This differs significantly from the APIs you've showed above, so it would be important to illustrate these here even if only a little bit.

heathermiller: What are the limitations of Sawzall and Pig? Why invest so much more effort on Hive? There have to be pretty significant differences between the tools. Point those differences out. You only need to invest 1-2 sentences doing so.

chapter/8/big-data.md
```diff
@@ -1,11 +1,473 @@
...
+Hive is built by Facebook to organize dataset in structured formats and still utilize the benefit of MapReduce framework. It has its own SQL-like language: HiveQL  {%cite thusoo2010hive --file big-data %} which is easy for anyone who understands SQL. Hive reduces code complexity and eliminates lots of boiler plate that would otherwise be an overhead with Java based MapReduce approach.  It has a component called *metastore* that are created and reused each time the table is referenced by HiveQL like the way traditional warehousing solutions do. The drawback to using Hive is programmers have to be familiar with basic techniques and best practices for running their Hive queries at maximum speed as it depends on the Hive optimizer. Hive requires developers  train the Hive optimizer for efficient optimization of their queries.
+
+Relational interface to big data is good, however, it doesn’t cater to users who want to perform
+
+- ETL to and from various semi or unstructured data sources.
+- advanced analytics like machine learning or graph processing.
```
heathermiller, line 156: Ok, I realize that this section isn't done. My previous comment still holds. Please make sure to compare/contrast these query language interfaces to MapReduce/Spark so we can clearly understand the benefits/limitations across each.

chapter/8/big-data.md
```diff
@@ -1,11 +1,473 @@
...
+
+***HiveSQL***: Hive query language consists of a subset of SQL along with some extensions. The language is very SQL-like and supports features like subqueries, joins, cartesian product, group by, aggregation, describe and more. MapReduce programs can also be used in Hive queries. A sample query using MapReduce would look like this:
+```
+FROM (
+    MAP inputdata USING 'python mapper.py' AS (word, count)
+    FROM inputtable
+    CLUSTER BY word
+    )
+    REDUCE word, count USING 'python reduce.py';
+```
```
heathermiller, line 187: Great example! 😄

chapter/8/big-data.md
```diff
@@ -1,11 +1,473 @@
...
+This query uses mapper.py for transforming inputdata into (word, count) pair, distributes data to reducers by hashing on word column (given by CLUSTER) and uses reduce.py.
+INSERT INTO, UPDATE, and DELETE are not supported which makes it easier to handle reader and writer concurrency.
```
heathermiller, line 189: ...?

chapter/8/big-data.md
```diff
@@ -1,11 +1,473 @@
...
+
+
+***Serialization/Deserialization***
+Hive implements the LazySerDe as the default SerDe interface. A SerDe is a combination of serialization and deserialization which helps developers instruct Hive on how their records should be processed. The Deserializer interface translates rows into internal objects lazily so that the cost of Deserialization of a column is incurred only when it is needed. The Serializer, however, converts a Java object into a format that Hive can write to HDFS or another supported system. Hive also provides a RegexSerDe which allows the use of regular expressions to parse columns out from a row.
+
+### 1.2.2 Pig Latin
+The goal of Pig Latin {% cite olston2008pig --file big-data%} is to attract experienced programmers to perform ad-hoc analysis on big data. Parallel database products provide a simple SQL query interface, which is good for non-programmers and simple tasks, but not in a style where experienced programmers would approach. Instead such programmers prefer to specify single steps and operate as a sequence.
```
heathermiller, line 196: If I wasn't familiar with Pig Late, I wouldn't understand from this paragraph what Pig Latin really is. I suspect you might have taken descriptions from the paper; the "goal is to attract experienced programmers to perform ad-hoc analysis on big data" point sticks out at me, but confuses me in this context. I would just remove this sentence.

Otherwise, you mean to say that Pig Latin provides two interfaces, both an SQL-like query language designed for non-programmers, and a functional-like programming model for experienced programmers. Rephrase to better make this point in this paragraph.

chapter/8/big-data.md
```diff
@@ -1,11 +1,473 @@
...
+big_groups = FILTER groups BY COUNT(good_urls)>106;
+output = FOREACH big_groups GENERATE
+            category, AVG(good_urls.pagerank);
+```
+
+*Interoperability* Pig Latin is designed to support ad-hoc data analysis, which means the input only requires a function to parse the content of files into tuples. This saves the time-consuming import step. While as for the output, Pig provides freedom to convert tuples into byte sequence where the format can be defined by users.  
+
+*Nested Data Model* Pig Latin has a flexible, fully nested data model, and allows complex, non-atomic data types such as set, map, and tuple to occur as fields of a table. The benefits include: closer to how programmer think; data can be stored in the same nested fashion to save recombining time; can have algebraic language; allow rich user defined functions.  
+
+*UDFs as First-Class Citizens* Pig Latin supports user-defined functions (UDFs) to support customized tasks for grouping, filtering, or per-tuple processing.
```
heathermiller, line 220: More detail here. Explain why this is desirable.

chapter/8/big-data.md
```diff
@@ -1,11 +1,473 @@
...
+A Dataframe API lets users to intermix both relational and procedural code with ease. Dataframe is a collection of schema based rows of data and named columns on which relational operations can be performed with optimized execution. Unlike a RDD, Dataframe allows developers to define structure for the data and can be related to tables in a relational database or R/Python’s Dataframe. Dataframe can be constructed from tables of external sources or existing native RDD’s. Dataframe is lazy and each object in it represents a logical plan which is not executed until an output operation like save or count is performed.
+Spark SQL supports all the major SQL data types including complex data types like arrays, maps and unions.
+Some of the Dataframe operations include projection (select), filter(where), join and aggregations(groupBy).
+Illustrated below is an example of relational operations on employees data frame to compute the number of female employees in each department.
+
+```
+employees.join(dept, employees("deptId") === dept("id")) .where(employees("gender") === "female") .groupBy(dept("id"), dept("name")) .agg(count("name"))
+```
```
heathermiller, line 238: Seems the syntax here is broken. I think you meant to do this:

```scala
employees.join(dept, employees("deptId") === dept("id")) 
         .where(employees("gender") === "female") 
         .groupBy(dept("id"), dept("name")) 
         .agg(count("name"))
```

chapter/8/big-data.md
```diff
@@ -1,11 +1,473 @@
...
+Several of these operators like  === for equality test, > for greater than, a rithmetic ones (+, -, etc) and aggregators transforms to a abstract syntax tree of the expression which can be passed to Catalyst for optimization.
+A cache() operation on the data frame helps Spark SQL store the data in memory so it can be used in iterative algorithms and for interactive queries. In case of Spark SQL, memory footprint is considerably less as it applies columnar compression schemes like dictionary encoding / run-length encoding.
+
+The DataFrame API also supports inline UDF definitions without complicated packaging and registration. Because UDFs and queries are both expressed in the same general purpose language (Python or Scala), users can use standard debugging tools.
+
+However, a DataFrame lacks type safety. In the above example, attributes are referred to by string names. Hence, it is not possible for the compiler to catch any errors. If attribute names are incorrect then the error will only detected at runtime, when the query plan is created.
```
heathermiller, line 244: > Hence, it is not possible for the compiler to catch any errors.

In addition to this point...

This is both very brittle and very verbose as well, because the user has to cast each row and column to specific types before they can do anything on them. Naturally this is very error-prone because you could accidentally choose the wrong index for a row/column and end up with a `ClassCastException`. This might be a worthwhile point to make here.

chapter/8/big-data.md
```diff
@@ -1,11 +1,473 @@
...
+Spark introduced a extension to Dataframe called ***Dataset*** to provide this compile type safety. It embraces object oriented style for programming and has an additional feature termed Encoders. Encoders translate between JVM representations (objects) and Spark’s internal binary format. Spark has built-in encoders which are very advanced in that they generate byte code to interact with off-heap data and provide on-demand access to individual attributes without having to de-serialize an entire object
+
+
+Winding up - we can compare SQL vs Dataframe vs Dataset as below :
+
+<figure class="main-container">
+  <img src="./sql-vs-dataframes-vs-datasets.png" alt="SQL vs Dataframe vs Dataset" />
+</figure>
+*Figure from the website :* https://databricks.com/blog/2016/07/14/a-tale-of-three-apache-spark-apis-rdds-dataframes-and-datasets.html
```
heathermiller, line 253: Thank you! Please make sure to also cite the source of any other figure you may have used.

chapter/8/big-data.md
```diff
@@ -1,11 +1,473 @@
...
 
-{% bibliography --file big-data %}
\ No newline at end of file
+### 1.3 Large-scale Parallelism on Graphs
+Map Reduce doesn’t scale easily and is highly inefficient for iterative / graph algorithms like page rank and machine learning algorithms. Iterative algorithms requires programmer to explicitly handle the intermediate results (writing to disks). Hence, every iteration requires reading the input file and writing the results to the disk resulting in high disk I/O which is a performance bottleneck for any batch processing system.
```
heathermiller, line 256: > is highly inefficient for iterative / graph algorithms

I wouldn't say that it's only inefficient. It's also simply difficult to use for iterative algorithms (in terms of the amount of boilerplate you have to write for each stage in the computation), in addition to the performance hit you get.

Point this out.

chapter/8/big-data.md
```diff
@@ -1,11 +1,473 @@
...
+
+Also graph algorithms require exchange of messages between vertices. In case of PageRank, every vertex requires the contributions from all its adjacent nodes to calculate its score. Map reduce currently lacks this model of message passing which makes it complex to reason about graph algorithms. One model that is commonly employed for implementing distributed graph processing is the graph parallel model.
+
+In the graph-parallel abstraction, a user-defined vertex program is instantiated concurrently for each vertex and interacts with adjacent vertex programs through messages or shared state. Each vertex program can read and modify its vertex property and in some cases adjacent vertex properties. When all vertex programs vote to halt the program terminates. Most systems adopt the bulk synchronous parallel model {% cite bulk-synchronous-model --file big-data%}.
+
+This model was introduced in 1980 to represent the hardware design features of parallel computers. It gained popularity as an alternative for map reduce since it addressed the above mentioned issues with map reduce<br />
```
heathermiller, line 262: > This model was introduced in 1980

Cite Valiant's 1980 paper on the BSP model here as well.

chapter/8/big-data.md
```diff
@@ -1,11 +1,473 @@
...
+ - Barrier synchronization synchs all the processors at the end of every superstep.
+
+A notable feature of the model is the complete control on data through communication between every processor at every superstep. Though similar to map reduce model, BSP preserves data in memory across supersteps and helps in reasoning iterative graph algorithms.
+
+The graph-parallel abstractions allow users to succinctly describe graph algorithms, and provide a runtime engine to execute these algorithms in a distributed nature. They simplify the design, implementation, and application of sophisticated graph algorithms to large-scale real-world problems. Each of these frameworks presents a different view of graph computation, tailored to an originating domain or family of graph algorithms. However, these frameworks fail to address the problems of data preprocessing and construction, favor snapshot recovery over fault tolerance and lack support from distributed data flow frameworks. The data-parallel systems are well suited to the task of graph construction, and are highly scalable. However, suffer from the very problems mentioned before for which the graph-parallel systems came into existence.
```
heathermiller, line 273: It seems like you  need to figure out how to better connect the BSP section with the GraphX section. Seems you're not quite finished with this section yet.

chapter/8/big-data.md
```diff
@@ -1,11 +1,473 @@
...
+GraphX {%cite xin2013graphx --file big-data%} is a new computation system which builds upon the Spark’s Resilient Distributed Dataset (RDD) to form a new abstraction Resilient Distributed Graph (RDG) to represent records and their relations as vertices and edges respectively. RDG’s leverage the RDD’s fault tolerance mechanism and expressivity.
+
+How does GraphX improve over the existing graph-parallel and data flow models ?
+The RDGs in GraphX provides a set of elegant and expressive computational primitives through which  many a graph parallel systems like Pregel, PowerGraph can be easily expressed with minimal lines of code. GraphX simplifies the process of graph ETL and analysis through new operations like filter, view and graph transformations. It minimizes communication and storage overhead.
+
+Similar to the data flow model, it GraphX away from the vertex centric view and adopts transformations on graphs yielding a new graph.
```
heathermiller, line 279: I can see that more work is needed on this section, so I will hold off on giving feedback here.

chapter/8/big-data.md
```diff
@@ -1,11 +1,473 @@
...
+
+***Why partitioning is important in graph computation systems ?***
```
heathermiller, line 281: I love these sections!! 😍

chapter/8/big-data.md
```diff
@@ -1,11 +1,473 @@
...
+
+***Why Edge-cuts are expensive ?***
+Edge-cuts for partitioning requires random assignment of vertices and edges across all the machines. hus the communication and storage overhead is proportional to the number of edges cut, and this makes balancing the number of cuts a priority. For most real-world graphs, constructing an optimal edge-cut is cost prohibitive, and most systems use random edge-cuts which achieve appropriate work balance, but nearly worst-case communication overhead.
+
+<figure class="main-container">
+  <img src="./vertex-cuts.png" alt="Vertex cuts" />
+</figure>
+*Figure from {%cite xin2013graphx --file big-data%}*
+
+***Vertex-cuts - GraphX’s solution to effective partitioning*** : An alternative approach which does the opposite of edge-cut — evenly assign edges to machines, but allow vertices to span multiple machines. The communication and storage overhead of a vertex-cut is directly proportional to the sum of the number of machines spanned by each vertex. Therefore, we can reduce communication overhead and ensure balanced computation by evenly assigning edges to machines in way that minimizes the number of machines spanned by each vertex.
```
heathermiller, line 297: I love these sections!! 😍

chapter/8/big-data.md
```diff
@@ -1,11 +1,473 @@
...
+- subgraph - Applies predicates to return a subgraph of the original graph by filtering all the vertices and edges that don't satisfy the vertices and edges predicates respectively.
+- mrTriplets (MapReduce triplet) - logical composition of triplets followed by map and reduceByKey. It is the building block of graph-parallel algorithms.
+
+## 2 Execution Models
+There are many possible implementations for those programming models. In this section, we will discuss about a few different execution models, how the above programming interfaces exploit them, the benefits and limitations of each design and so on. MapReduce, its variants and Spark all use the master/workers model (section 2.1), where the master is responsible for managing data and dynamically scheduling tasks to workers. The master monitors workers' status, and when failure happens, master will reschedule the task to another idle worker. The fault-tolerance is guaranteed by persistence of data in MapReduce versus lineage(for recomputation) in Spark.
+
+
+
+### 2.1 Master/Worker model
```
heathermiller, line 323: This is a little confusing because above you say that MapReduce and Spark both have the same underlying model of execution; the Master/Worker model. While that might be true macroscopically, of course the whole disk vs memory thing is a big difference.

Therefore, in the introduction paragraph, I would make the argument that master/worker is adopted by both MapReduce and Spark, but realized in very different ways. Use the explanation you gave about fault tolerance that you gave above: 

> The fault-tolerance is guaranteed by persistence of data in MapReduce versus lineage(for recomputation) in Spark.

Then, I would name the subsections something different. Because after your intro, I'm left confused because section "2.1 Master/Worker model" is really all about MapReduce, not Spark as well (as one might assume from your intro paragraphs)

Rename this section to something else instead.

chapter/8/big-data.md
```diff
@@ -1,11 +1,473 @@
...
+1. Split data. The input files are split into *M* pieces;
+2. Copy processes. The user program create a master process and the workers. The master picks idle workers to do either map or reduce task;
+3. Map. The map worker reads corresponding splits and passes to the map function. The generated intermediate key/value pairs are buffered in memory;
+4. Partition. The buffered pairs are written to local disk and partitioned to *R* regions periodically. Then the locations are passed back to the master;
+5. Shuffle. The reduce worker reads from the local disks and groups together all occurrences of the same key together;
+6. Reduce. The reduce worker iterates over the grouped intermediate data and calls reduce function on each key and its set of values. The worker appends the output to a final output file;
+7. Wake up. When all tasks finish, the master wakes up the user program.
+
+<figure class="fullwidth">
+  <img src="{{ site.baseurl }}/resources/img/mapreduce-execution.png" alt="MapReduce Execution Overview" />
```
heathermiller, line 334: Nice! 😄

chapter/8/big-data.md
```diff
@@ -1,11 +1,473 @@
...
+
+There are some practices in this paper that make the model work very well in Google, one of them is **backup tasks**: when a MapReduce operation is close to completion, the master schedules backup executions of the remaining in-progress tasks ("straggler"). The task is marked as completed whenever either the primary or the backup execution completes.
+In the paper, the authors measure the performance of MapReduce on two computations running on a large cluster of machines. One computation *grep* through approximately 1TB of data. The other computation *sort* approximately 1TB of data. Both computations take in the order of a hundred seconds. In addition, the backup tasks do help largely reduce execution time. In the experiment where 200 out of 1746 tasks were intentionally killed, the scheduler was able to recover quickly and finish the whole computation for just a 5% increased time.  
+Overall, the performance is very good for conceptually unrelated computations.
```
heathermiller, line 342: This performance analysis section seems a bit out of place here, because it is impossible to compare it to the rest of the paper. I leave it up to you, but it might be better to get rid of it (you don't have to, just my 2c)

chapter/8/big-data.md
```diff
@@ -1,11 +1,473 @@
...
+
+### 2.2 Spark execution model
+
+<figure class="main-container">
+  <img src="./cluster-overview.png" alt="MapReduce Execution Overview" />
+</figure>
+
+The Spark driver defines SparkContext which is the entry point for any job that defines the environment/configuration and the dependencies of the submitted job. It connects to the cluster manager and requests resources for further execution of the jobs.
+The cluster manager manages and allocates the required system resources to the Spark jobs. Furthermore, it coordinates and keeps track of the live/dead nodes in a cluster. It enables the execution of jobs submitted by the driver on the worker nodes (also called Spark workers) and finally tracks and shows the status of various jobs running by the worker nodes.
+A Spark worker executes the business logic submitted by the Spark driver. Spark workers are abstracted and are allocated dynamically by the cluster manager to the Spark driver for the execution of submitted jobs. The driver will listen for and accept incoming connections from its executors throughout its lifetime.
```
heathermiller, line 353: > A Spark worker executes the business logic submitted by the Spark driver.

I would say:

>...submitted by the user by way of the Spark driver

chapter/8/big-data.md
```diff
@@ -1,11 +1,473 @@
...
+
+***How are persistent RDD’s memory managed ?***
+
+Persistent RDDs are stored in memory as java objects (for performance) or in memory as serialized data (for less memory usage at cost of performance) or on disk. If the worker runs out of memory upon creation of a new RDD, LRU policy is applied to evict the least recently accessed RDD unless its same as the new RDD. In that case, the old RDD is excluded from eviction given the fact that it may be reused again in future. Long lineage chains involving wide dependencies are checkpointed to reduce the time in recovering a RDD. However, since RDDs are read-only, checkpointing is still ok since consistency is not a concern and there is no overhead to manage the consistency as is seen in distributed shared memory.
```
heathermiller, line 359: For the beginners that might read this, you should do this:

> a Least Recently Used (LRU) policy is applied

chapter/8/big-data.md
```diff
@@ -1,11 +1,473 @@
...
+ - Transformation of AST to DAG (Directed acyclic graph) : In this phase it generates logical plan and does a compile type checking. Logical plan is generated using the metadata (stored in Metastore) information of the required tables. It can flag errors if any issues found during the type checking.
+
+ - Optimization : Optimization forms the core of any declarative interface. In case of Hive, optimization happens through chains of transformation of DAG. A transformation could include even a user defined optimization and it applies an action on the DAG only if a rule is satisfied. Every node in the DAG implements a special interface called as Node interface which makes it easy for the manipulation of the operator DAG using other interfaces like GraphWalker, Dispatcher, Rule and Processor. Hence, by transformation, we mean walking through a DAG and for every Node we encounter we perform a Rule satisfiability check. If a Rule is satisfied, a corresponding processor is invoked. A Dispatcher maintains a list of Rule to Processor mappings.
+
+<figure class="main-container">
+  <img src="./Hive-transformation.png" alt="Hive transformation" />
```
heathermiller, line 377: This figure is broken or otherwise unreadable when the page is generated.

chapter/8/big-data.md
```diff
@@ -1,11 +1,473 @@
...
+</figure>
+*Figure to depict the transformation flow during optimization, from:* %cite thusoo2010hive --file big-data %}
```
heathermiller, line 379: Broken citation

chapter/8/big-data.md
```diff
@@ -1,11 +1,473 @@
...
+
+ Some of the important transformations are :
```
heathermiller, line 381: Seems like you're not quite done with this section. Need to polish it a bit and tie it in better with the rest above.

chapter/8/big-data.md
```diff
@@ -1,11 +1,473 @@
...
+
+Catalyst leverages the Scala’s functional language features like pattern matching and runtime meta programming to allow developers to concisely specify complex relational optimizations.
+
+Catalyst includes both rule-based and cost-based optimization. It is extensible to include new optimization techniques and features to Spark SQL and also let developers provide data source specific rules.
+Catalyst executes the rules on its data type Tree - a composition of node objects where each node has a node type (subclasses of TreeNode class in Scala) and zero or more children. Node objects are immutable and can be manipulated. The transform method of a Tree applies pattern matching to match a subset of all possible input trees on which the optimization rules needs to be applied.
+
+Hence, in Spark SQL, transformation of user queries happens in four phases :
+
+<figure class="main-container">
+  <img src="./sparksql-data-flow.jpg" alt="SparkSQL optimization plan Overview" />
```
heathermiller, line 415: Make sure to cite the source of this figure!

chapter/8/big-data.md
```diff
@@ -1,11 +1,473 @@
...
+
+
+***Code Generation :*** The final phase generates the Java byte code that should run on each machine.Catalyst transforms the Tree which is an expression in SQL to an AST for Scala code to evaluate, compile and run the generated code. A special scala feature namely quasiquotes aid in the construction of abstract syntax tree(AST).
+
+
+## 3. Big Data Ecosystem
+*Hadoop Ecosystem*  
+
+Apache Hadoop is an open-sourced framework that supports distributed processing of large dataset. It involves a long list of projects that you can find in this table https://hadoopecosystemtable.github.io/. In this section, it is also important to understand the key players in the system, namely two parts: the Hadoop Distributed File System (HDFS) and the open-sourced implementation of MapReduce model - Hadoop.
```
heathermiller, line 431: > projects that you can find in this table https://hadoopecosystemtable.github.io/. 

Links in markdown are done like this:
```
[link text](https://hadoopecosystemtable.github.io/)
```

heathermiller: > It involves a long list of projects that you can find in this table

I would say,

> it involves dozens of projects, all of which are listed here

^ Add the link to the "here" text.

chapter/8/big-data.md
```diff
@@ -1,11 +1,473 @@
...
+
+
+HDFS forms the data management layer, which is a distributed file system designed to provide reliable, scalable storage across large clusters of unreliable commodity machines. The idea was inspired by GFS{%cite ghemawat2003google --file big-data%}. Unlike closed GFS, HDFS is open-sourced and provides various libraries and interfaces to support different file systems, like S3, KFS etc.
+
+To satisfy different needs, big companies like Facebook and Yahoo developed additional tools. Facebook's Hive, as a warehouse system, can provide more declarative programming interface and translate to Hadoop jobs. Yahoo's Pig platform is an ad-hoc analysis tool that can structurize HDFS objects and support operations like grouping, joining and filtering.   
+
+
+***Spark Ecosystem***
+
+Apache Spark's rich-ecosystem constitutes of third party libraries like Mesos{%cite hindman2011mesos --file big-data%}/Yarn{%cite vavilapalli2013apache --file big-data%} and several major components that have been already discussed in this article like Spark-core, SparkSQL, GraphX.
```
heathermiller, line 446: Don't forget to point out that Spark is a part of the Hadoop ecosystem! At the very least, the most popular way for Spark to read in data is via HDFS.

chapter/8/big-data.md
```diff
@@ -1,11 +1,473 @@
...
+
+
+*Apache Mesos*
+
+Apache Mesos{%cite hindman2011mesos --file big-data%} is an open source cluster/resource manager developed at the University of California, Berkley and used by  companies such  as Twitter, Airbnb, Netflix etc. for handling workloads in a distributed environment through dynamic resource sharing and isolation. It aids in the deployment and management of applications in large-scale clustered environments. Mesos abstracts node allocation by combining the existing resources of the machines/nodes in a cluster into a single pool and enabling fault-tolerant elastic distributed systems. Variety of workloads can utilize the nodes from this single pool voiding the need of allocating specific machines for different workloads. Mesos is highly scalable, achieves fault tolerance through Apache Zookeeper and is a efficient CPU and memory-aware resource scheduler.
```
heathermiller, line 460: > ...for handling workloads in a distributed environment through dynamic resource sharing and isolation.

Isn't Mesos also designed for heterogeneous clusters? That is, you should be able to run all kinds of other platforms/runtimes alongside of Hadoop/Spark?

Point this out.

heathermiller (commented):
This is really nice guys! You've come a **really long way**! There's not a whole lot of work left to do on this. Small things that you've got to do for the final revision are left as inline comments.

Biggish things that you need to resolve in the final revision:

- Invest more effort in the Dryad section, including adding examples (details inline)
- Invest more effort in the introduction to the query language interfaces to MapReduce/Spark, "1.2 Querying: declarative interfaces" including adding examples (details inline)
- Run your text through a spellchecker/grammar checker. It's easy to install this in Sublime Text or Atom or whatever you use in just a few seconds. After you pass me your final revision, I'm happy to help out with wording, grammar, etc, but the spellcheck/grammar check could save everyone a bunch of time (there are a bunch of little mistakes)
- Revisit the first paragraph (introduction paragraph at the top of the chapter) when you're finished with the chapter. It can be improved, it's a bit salesy, and I imagine you'll have a better idea how to pitch your chapter once you've finished it.
- Pass over and polish _a few times_. Many sentences are a bit rushed, sentences are repeated, some sentences end abruptly. Your structure/arguments are all great, but the entire chapter just needs polish. There are lots of rough edges.

That's really it! You guys are really close to being finished! You need just a day or so more of working on it and you'll be done :)

You should be proud of yourselves. This is a really nice article that I think is _very_ useful to both developers & researchers! 😄

heathermiller:
Hi guys, I'm merging this now (even though it's based on @cnnrznn's branch. I'll help @cnnrznn deal with any merge conflicts that emerge from me merging this...)

**Please make a new PR with the above changes addressed!**

chapter/8/big-data.md
msabhi: Here I was comparing relational only interfaces (like Hive) to relational + procedural interfaces (PigLatin/SparkSQL)...Do you think I should retain this comparison?
regardless of including above comparison or not, I will still add lines to compare against map reduce/spark.

heathermiller: I think you could make that point, sure. But you don't have to invest a lot of effort or have separate subsections for "relational only" and "relational + procedural". So sure, make the point, but you can be short/concise about it. Hope that makes sense.

msabhi: Sure professor.

chapter/8/big-data.md
msabhi: So, I realized that BSP is one of graph-parallel models. Hence, before explaining BSP, I connect it to graph parallel model. I mention it as the most widely used graph-parallel abstraction and then explain BSP. Hence, I connected graph-parallel with the graphx later. Do you think its fine?

msabhi: In the graph-parallel abstraction, a user-defined vertex program is instantiated concurrently for each vertex and interacts with adjacent vertex programs through messages or shared state. Each vertex program can read and modify its vertex property and in some cases adjacent vertex properties. When all vertex programs vote to halt the program terminates. Most systems adopt the **bulk synchronous parallel model** {% cite bulk-synchronous-model --file big-data%}.

This model was introduced in 1980 to represent the hardware design features of parallel computers. It gained popularity as an alternative for map reduce......

heathermiller: Oops I missed the notification for this comment. Yes it looks good!

PR #5, opened 2016-12-03
Message Passing Chapter

natdempk:
Working branch/PR for the message passing chapter.

Working outline and some research can be found here: https://docs.google.com/document/d/1zfxBMJRSGOD50gh3Z04kP4Tl1Ic-ScHBaUvgTe0_29Q/edit?usp=sharing


heathermiller:
Crazy suggestion that you're probably going to kill me for... What about drawing diagrams to try and differentiate the different actor models? I think it could *really* help your presentation.

natdempk:
@heathermiller I'm actually not opposed to the idea. Many of the sources I've read, especially the stuff like blog posts and that reactive systems ebook that are targeted towards business use utilize diagrams. That said, maybe we could meet and talk specific areas or topics where visualizations might be especially effective? I have some ideas but it would be good to be on the same page.

heathermiller:
Yeah, of course I don't think you should put it everywhere. But stuff like E desperately needs a diagram to be understood, I think. But yeah, let's meet. Name a day/time, I'm free whenever.

natdempk:
Tomorrow (12/5) at 1:00 PM work?

heathermiller:
Yup, sure. See you then.

heathermiller:
Merging this now so it's up on the site. Though I'll provide feedback in this (closed) PR.

chapter/3/message-passing.md
```diff
@@ -1,11 +1,250 @@
...
 
-{% bibliography --file message-passing %}
\ No newline at end of file
+In the field of message passing programming models, it is not only important to consider recent state of the art research, but additionally the historic initial papers on message passing and the actor model that are the roots of the programming models described in newer papers. It is enlightening to see which aspects of the models have stuck around, and many of the newer papers reference and address deficiencies present in older papers. There have been plenty of programing languages designed around message passing, especially those focused on the actor model of programming and organizing units of computation.
+
+In this chapter I describe four different actor models: classic actors, process-based actors, communicating event-loops, and active objects. I attempt to highlight historic and modern languages that exemplify these models, as well as the philosophies and tradeoffs that programmers need to be aware of to understand these models.
```
heathermiller, line 13: "In this chapter I describe four different actor models" -> "In this chapter I describe the four primary variants of the actor model"

heathermiller: "to understand these models." -> "to understand and make best use of these models."

chapter/3/message-passing.md
```diff
@@ -1,11 +1,250 @@
...
+
+Actor programming models are continuing to develop and become more robust, as some of the recently published papers and systems in the field show. There are a few robust industrial-strength actor systems that are being used to power massive scalable distributed systems. There are a couple of different approaches to building actor frameworks that are detailed later in the chapter.
```
heathermiller, line 15: > Actor programming models are continuing to develop and become more robust, as some of the recently published papers and systems in the field show.

I'd say instead,

> Despite the actor model's originating as far back as the 1970s, it is still being developed and being incorporated into the programming languages of today, as many recently published papers and systems in the field demonstrate.

heathermiller: > There are a few robust industrial-strength actor systems that are being used to power massive scalable distributed systems.

Weak statement unless you cite such a system. Either cite some systems and mention them by name, or omit this.

heathermiller: > There are a couple of different approaches to building actor frameworks that are detailed later in the chapter.

Omit this or improve it. Do you mean to say industrial frameworks for monitoring and handling fault-tolerance in industrial systems? If so, say so. Else this as-is, this sentence is a bit too vague.

chapter/3/message-passing.md
```diff
@@ -1,11 +1,250 @@
...
+
+I think an important framing for the models and sources presented is “Why message passing, and specifically why the actor model?” There are a number of distributed programming models, so why was this one so important when it was initially proposed? What are the advantages of it for the programmer? Why has it facilitated advanced languages, systems, and libraries that are widely used today? The broad advantages of the actor model are around isolation of state, scalability, and simplifying the programmer's ability to reason about their system.
```
heathermiller, line 17: > ~~I think a~~An important framing for the models and sources presented is

(Remove  uncertainty. You're the person teaching the reader here)

heathermiller: I would say instead:

> An important framing for the **actor models**~~and sources~~ presented is **in the question** “Why message passing, and specifically why the actor model?”

heathermiller: > There are a number of distributed programming models, so why was this one so important when it was initially proposed?

I'd say instead:

> Given the vast number of distributed programming models out there, one might ask, why this one was so important when it was initially proposed?

heathermiller, line 17: > The broad advantages of the actor model are around isolation of state, scalability, and simplifying the programmer's ability to reason about their system.

You might want to rephrase this into something that entices the reader to continue reading the rest of the paper. E.g., 

> As we'll see throughout this chapter, some of the broadest advantages of the actor model include isolation of state managed by the given actor, scalability, and simplifying the programmer's ability to reason about their system.

chapter/3/message-passing.md
```diff
@@ -1,11 +1,250 @@
...
+
+# Original proposal of the actor model
+
+The actor model was originally proposed in _A Universal Modular ACTOR Formalism for Artificial Intelligence_ in 1973 as a method of computation for artificial intelligence research. The original goal of the model was to model parallel computation in communication in a way that could be safely distributed concurrently across workstations. The paper makes few presumptions about implementation details, instead defining the high-level message passing communication model.
```
heathermiller, line 21: > The actor model was originally proposed in _A Universal Modular ACTOR Formalism for Artificial Intelligence_ in 1973 

Citation here.

chapter/3/message-passing.md
```diff
@@ -1,11 +1,250 @@
...
+
+They define actors as independent units of computation with isolated state. These units can send messages to one another, and have a mailbox which contains messages they have received. These messages are of the form:
```
heathermiller, line 23: As silly as this sounds, it might help to make a bulleted list here. Just so people can't quickly gloss over this sentence describing the essential makeup of an actor. E.g., 

> Actors are defined as independent units of computation with isolated state. These units have two core characteristics:
> - they can send messages to one another, and,
> - they have a mailbox which contains messages that they have received.
>
> Messages are of the form:

chapter/3/message-passing.md
```diff
@@ -1,11 +1,250 @@
...
+
+```
+(request: <message-to-target>
+ reply-to: <reference-to-messenger>)
+```
+
+Actors attempt to process messages from their mailboxes by matching their `request` field sequentially against patterns or rules which can be specific values or logical statements. When a pattern is matched, computation occurs and the result of that computation is implicitly returned to the reference in the message's `reply-to` field. This is a type of continuation, where the continuation is the message to another actor. These messages are one-way and make no claims about whether a message will ever be received in response. This model is limited compared to many of the others, but the early ideas of taking advantage of distribution of processing power to enable greater parallel computation are there.
```
heathermiller, line 30: > These messages are one-way and make no claims about whether a message will ever be received in response.

I'd say instead,

> These messages are one-way and, there are no guarantees that a message will ever be received in response.

heathermiller: > This model is limited compared to many of the others, but the early ideas of taking advantage of distribution of processing power to enable greater parallel computation are there.

Too vague. Do you mean this specific variant of the actor model? Be explicit here that you're talking about the originally-proposed variant of the actor model.

chapter/3/message-passing.md
```diff
@@ -1,11 +1,250 @@
...
+
+One interesting thing to note is that this original paper talks about actors in the context of hardware. They mention actors as almost another machine architecture. This paper describes the concepts of an "actor machine" and a "hardware actor" as the context for the actor model, which is totally different from the way we think about modern actors as abstracting away a lot of the hardware details we don't want to deal with. This concept reminds me of something like a Lisp machine, but built to specially utilize the actor model of computation for artificial intelligence.
```
heathermiller, line 32: > One interesting thing to note is that this original paper talks about actors in the context of hardware.

I'd say instead,

> Interestingly, the original paper introducing the actor model does so in the context of hardware.

heathermiller: > This concept reminds me of something like a Lisp machine, but built to specially utilize the actor model of computation for artificial intelligence.

I'd say,

> This concept is reminiscent of something like a Lisp machine, though specially built to utilize the actor model of computation for artificial intelligence.

chapter/3/message-passing.md
```diff
@@ -1,11 +1,250 @@
...
+
+# Classic actor model
+
+The classic actor model was formalized as a unit of computation in Agha's _Concurrent Object-Oriented Programming_. The classic actor expands on the original proposal of actors, keeping the ideas of asynchronous communication through messages between isolated units of computation and state. The classic actor is formalized as the following primitive operations:
```
heathermiller, line 36: > The classic actor model was formalized as a unit of computation in Agha's _Concurrent Object-Oriented Programming_.

Citation here.

heathermiller: > The classic actor is formalized as the following primitive operations:

I'd say instead,

> The classic actor contains the following primitive operations:

chapter/3/message-passing.md
```diff
@@ -1,11 +1,250 @@
...
+
+* `create`: create an actor from a behavior description and a set of parameters, including other existing actors
+* `send`: send a message to another actor
+* `become`: have an actor replace their behavior with a new one
+
+As originally described, classic actors communicate by asynchronous message passing. They are a primitive independent unit of computation which can be used to build higher-level abstractions for concurrent programming. Actors are uniquely addressable, and have their own independent mailboxes or message queues. State changes using the classic actor model are specified and aggregated using the `become` operation. Each time an actor processes a communication it computes a behavior in response to the next type of communication it expects to process. A `become` operation's argument is another named behavior with some state to pass to that named behavior.
```
heathermiller, line 42: > As originally described, classic actors communicate by asynchronous message passing.

I'd say instead,

> As in the original actor model, classic actors communicate by asynchronous message passing.

heathermiller: > Each time an actor processes a communication

"communication" -> "message"

heathermiller: > in response to the next type of communication it expects to process

"communication" -> "message"

heathermiller: > A `become` operation's argument is another named behavior with some state to pass to that named behavior.

By "another named behavior," you mean a continuation, right? Maybe "continuation" is a better word than "behavior" here?

E.g, 

> A `become` operation's argument is a named continuation, `b`, representing behavior that the actor should be updated with along with some state that should be passed to `b`.

chapter/3/message-passing.md
```diff
@@ -1,11 +1,250 @@
...
+
+For purely functional actors the new behavior would be identical to the original. For more complex actors however, this enables the aggregation of state changes at a higher level of granularity than something like a variable assignment. This enables flexibility in the behavior of an actor over time in response to the actions of other actors in the system. Additionally, this isolation changes the level at which one analyzes a system, freeing the programmer from worrying about interference during state changes.
```
heathermiller, line 44: > For purely functional actors the new behavior would be identical to the original. For more complex actors however, this enables the aggregation of state changes at a higher level of granularity than something like a variable assignment.

^ This is a bit confusing.

chapter/3/message-passing.md
```diff
@@ -1,11 +1,250 @@
...
+
+If you squint a little, this actor definition sounds similar to Alan Kay’s original definition of Object Oriented programming. This definition describes a system where objects have a behavior, their own memory, and communicate by sending and receiving messages that may contain other objects or simply trigger actions. Kay's ideas sound closer to what we consider the actor model today, and less like what we consider object-oriented programming. The focus is on designing the messaging and communications that dictate how objects interact.
```
heathermiller, line 46: > If you squint a little, this actor definition sounds similar to Alan Kay’s original definition of Object Oriented programming. 

I think there's a nice quote by Alan Kay that you can probably easily find and throw in here.

heathermiller: > Kay's ideas sound closer to what we consider the actor model today, and less like what we consider object-oriented programming.

🎉

heathermiller: > The focus is on designing the messaging and communications that dictate how objects interact.

Expand a little, it's getting vague again

> That is, Kay's focus in this description is on designing the messaging and communications that dictate how objects interact.

chapter/3/message-passing.md
```diff
@@ -1,11 +1,250 @@
...
+
+TODO: transition
+
+## Concurrent Object-Oriented Programming (1990)
+
+This is the seminal paper for the classic actor model, as it offers classic actors as a natural solution to solving problems at the intersection of two trends in computing: increased distributed computing resources and the rising popularity of object-oriented programming. The paper defines common patterns of parallelism: pipeline concurrency, divide and conquer, and cooperative problem solving. It then focuses on how the actor model can be used to solve these problems in an object-oriented style, and some of the challenges that arise with distributed actors and objects, as well as strategies and tradeoffs for communication and reasoning about behaviors.
```
heathermiller, line 52: > This is the seminal paper for the classic actor model,

I'd start this paragraph differently:

One could say that the renaissance of actor models in mainstream program began with seminal paper, _Concurrent Object-Oriented Programming_ (citation), as it offers classic actors as a natural solution to solving problems at the intersection of two trends in computing; increased distributed computing resources and the rising popularity of object-oriented programming.

chapter/3/message-passing.md
```diff
@@ -1,11 +1,250 @@
...
+
+This paper looks at a lot of systems and languages that are implementing solutions in this space, and starts to actually identify some of the programmer-centric advantages of actors. One of the core languages used for examples in the paper is Rosette, but the paper largely focuses on the potential and benefits of the model. Agha claims the benefits of using objects stem from a separation of concerns. "By separating the specification of what is done (the abstraction) from how it is done (the implementation), the concept of objects provides modularity necessary for programming in the large. It turns out that concurrency is a natural consequence of the concept of objects." Splitting concerns into multiple pieces allows for the programmer to have an easier time reasoning about the behavior of the program. It also allows the programmer to use more flexible abstractions in their programs, as Agha states. "It is important to note that the actor languages give special emphasis to developing flexible program structures which simplify reasoning about programs." This flexibility turns out to be a highly discussed advantage which continues to be touted in modern actor systems.
```
heathermiller, line 54: > and starts to ~~actually~~ identify

heathermiller: I'd say:

> starts to identify some of the advantages from the perspective of programmers of programming with actors.

heathermiller: > examples in the paper is Rosette

Citation.

heathermiller: Make this a blockquote (see README)

> "By separating the specification of what is done (the abstraction) from how it is done (the implementation), the concept of objects provides modularity necessary for programming in the large. It turns out that concurrency is a natural consequence of the concept of objects."

heathermiller: Make this a blockquote too (see README) 

> "It is important to note that the actor languages give special emphasis to developing flexible program structures which simplify reasoning about programs."

chapter/3/message-passing.md
```diff
@@ -1,11 +1,250 @@
...
+
+## Rosette
+
+Rosette was both a language for concurrent object-oriented programming of actors, as well as a runtime system for managing the usage of and access to resources by those actors. Rosette is mentioned throughout Agha's _Concurrent Object-Oriented Programming_, and the code examples given in the paper are written in Rosette. Agha is even an author on the Rosette paper, so its clear that Rosette is foundational to the classic actor model. It seems to be a language which almost defines what the classic actor model looks like in the context of concurrent object-oriented programming.
```
heathermiller, line 58: > Rosette

Citation.

heathermiller: > throughout Agha's _Concurrent Object-Oriented Programming_,

Citation.

chapter/3/message-passing.md
```diff
@@ -1,11 +1,250 @@
...
+
+The motivation behind Rosette was to provide strategies for dealing with problems like search, where the programmer needs a means to control how resources are allocated to sub-computations to optimize performance in the face of combinatorial explosion. This supports the use of concurrency in solving computationally intensive problems whose structure is not statically defined, but rather depends on some heuristic to return results. Rosette has an architecture which uses actors in two distinct ways. They describe two different layers with different responsibilities:
```
heathermiller, line 60: > The motivation behind Rosette was to provide strategies for dealing with problems like search, where the programmer needs a means to control how resources are allocated to sub-computations to optimize performance in the face of combinatorial explosion.

Can you elaborate a bit more on the search problem, just so the point about the sub-computations isn't lost? Without a bit more detail, this is too abstract to understand I think. Needs just a subsentence or so.

chapter/3/message-passing.md
```diff
@@ -1,11 +1,250 @@
...
+
+Akka provides a programming interface with both Java and Scala bindings for actors which looks similar to Scala Actors, but has different semantics in how it processes messages. Akka's `receive` operation defines a global message handler which doesn't block on the receipt of no matching messages, and is instead only triggered when a matching message can be processed. It also will not leave a message in an actor's mailbox if there is no matching patter to handle the message. The message will simply be discarded an an event will be published to the system. Akka's interface also provides stronger encapsulation to avoid exposing direct references to actors. To some degree this fixes problems in Scala Actors where public methods could be called on actors, breaking many of the guarantees programmers expect from message-passing. This system is not perfect, but in most cases it limits the programmer to simply sending messages to an actor using a limited interface.
```
heathermiller, line 79: Typo. "patter" -> "pattern"

heathermiller: > Akka's interface also provides stronger encapsulation to avoid exposing direct references to actors.

Can you further elaborate or show an example?

chapter/3/message-passing.md
```diff
@@ -1,11 +1,250 @@
...
+
+Process-based actors are defined as a computation which runs from start to completion, rather than the classic actor model, which defines an actor almost as a state machine of behaviors and the logic to transition between those. Similar state-machine like behavior transitions are possible through recursion with process-based actors, but programming them feels fundamentally different than using the previously described `become` statement.
+
+These actors use a `receive` primitive to specify messages that an actor can receive during a given state/point in time. `receive` statements have some notion of defining acceptable messages, usually based on patterns, conditionals or types. If a message is matched, corresponding code is evaluated, but otherwise the actor simply blocks until it gets a message that it knows how to handle. Depending on the language implementation `receive` might specify an explicit message type or perform some pattern matching on message values.
+
+## Erlang
```
heathermiller, line 93: Maybe at the end of this section it would be wise to briefly mention the ErlangVM? Since it has a special garbage collector, for example. This is of course unlike any other language.

chapter/3/message-passing.md
```diff
@@ -1,11 +1,250 @@
...
+
+The realization of this model depends on efficiently multiplexing actors to threads. This technique was originally developed in Scala actors, and later was adopted by Akka. This integration allows for Actors to invoke methods that block the underlying thread in a way that doesn't prevent actors from making process. This is important to consider in an event-driven system where handlers are executed on a thread pool, because the underlying event-handlers can't block threads without risking thread pool starvation. The end result here is that Scala Actors enabled a new lightweight concurrency primitive on the JVM, with enhancements over Erlang's model. In addition to the more natural abstraction, the Erlang model was further enhanced with Scala's type system and advanced pattern-matching capabilities.
```
heathermiller, line 119: Pattern-matching capabilities true. Though "type system" other than that, not so sure. 

One thing is that Scala Actors are of type `Any => Unit`, so they're essentially untyped. They receive literally any type, match on it, and since they're side-effecting, they return `Unit`. Might be cool to mention that here since you start talking about typed channels next in the Cloud Haskell section.

chapter/3/message-passing.md
```diff
@@ -1,11 +1,250 @@
...
+
+## Cloud Haskell
+
+Cloud Haskell is an extension or domain specific language of Haskell which essentially implements an enhanced version of the computational message-passing model of Erlang in Haskell. It enhances Erlang's model with advantages from Haskell's model of functional programming in the form of purity, types, and monads. Cloud Haskell enables the use of pure functions for remote computation, which means that these functions are idempotent and can be restarted or run elsewhere in the case of failure without worrying about side-effects or undo mechanisms. This alone isn't so different from Erlang, which operates on immutable data in the context of isolated memory.
```
heathermiller, line 123: > Cloud Haskell is an extension or domain specific language of Haskell which...

I'd just say extension. (Though this is debatable)

chapter/3/message-passing.md
```diff
@@ -1,11 +1,250 @@
...
+
+# Communicating event-loops
```
heathermiller, line 127: Fun fact: Akka was ported to JavaScript... https://github.com/unicredit/akka.js/

chapter/3/message-passing.md
```diff
@@ -1,11 +1,250 @@
...
+
+The motivation for this referencing model comes from wanting to work at a finer-grained level of references than a traditional actor exposes. The simplest example is that you want to ensure that another actor in your system can read a value, but can't write to it. How do you do that within another actor model? You might imagine creating a read-only variant of an actor which doesn't expose a write message type, or proxies only `read` messages to another actor which supports both `read` and `write` operations. In E because you are handing out object references, you would simply only pass around references to a `read` method, and you don't have to worry about other actors in your system being able to write values. These finer-grained references make reasoning about state guarantees easier because you are no longer exposing references to an entire actor, but instead the granular capabilities of the actor.
+
+TODO: write more here, maybe something around promise pipelining and partial failure? implications of different types of communication? maybe mention some of the points that inspire some aspects of modern actors?
```
heathermiller, line 144: Don't worry about making the promise pipelining point. (I think Sam is going to try to make it since he's talking about new language implementations vs language extensions. Bottomline: new language designs tend to inspire language extensions, much like what happened in the end for E. Promise pipelining in E inspired promises in JavaScript, but of course no one really uses E.)

Partial failure or implications of different types of communication would be cool.

chapter/3/message-passing.md
```diff
@@ -1,11 +1,250 @@
...
+
+# Communicating event-loops
+
+The communicating event-loop model was introduced in the E language, and is one that aims to change the level of granularity at which communication happens within an actor-based system. The previously described actor systems organize communication at the actor level, while the communicating event model puts communication between actors in the context of actions on objects within those actors. The overall messages still reference higher-level actors, but those messages refer to more granular actions within an actor's state.
+
+## E Language
+
+The E language implements a model which is closer to imperative object-oriented programming. Within a single actor-like node of computation called a "vat" many objects are contained. This vat contains not just objects, but a mailbox for all of the objects inside, as well as a call stack for methods on those objects. There is a shared message queue and event-loop that acts as one abstraction barrier for computation across actors. The actual references to objects within a vat are used for addressing communication and computation across actors and operate at a different level of abstraction.
```
heathermiller, line 133: Strongly suggest you show a diagram or two from the Concurrency Among Strangers paper (cite source)

chapter/3/message-passing.md
```diff
@@ -1,11 +1,250 @@
...
+
+* _Near reference_: This is a reference only possible between two objects in the same vat. These expose both synchronous immediate-calls and asynchronous eventual-sends.
+* _Eventual reference_: This is a reference which crosses vat boundaries, and only exposes asynchronous eventual-sends, not synchronous immediate-calls.
+
+The difference in semantics between the two types of references means that only objects within the same vat are granted synchronous access to one another. The most an eventual reference can do is asynchronously send and queue a message for processing at some unspecified point in the future. This means that within the execution of a vat, a degree of temporal isolation can be defined between the objects and communications within the vat, and the communications to and from other vats.
+
```
heathermiller, line 141: Please please put a code example here.

chapter/3/message-passing.md
```diff
@@ -1,11 +1,250 @@
...
+Transactions in Orleans are a way to causally reason about the different instances of actors that are involved in a computation. Because in this model computation happens in response to a single outside request, a given actor's chain of computation via. associated actors always contains a single instantiation of each actor. These causal chain of instantiations is treated as a single transaction. At reconciliation time Orleans uses these transactions, along with current instantiation state to reconcile to a consistent state.
+
+All of this is a longwinded way of saying that Orleans' programmer-centric contributions are that it separates the concerns of running and managing actor lifecycles from the concerns of how data flows throughout your distributed system. It does this is a fault-tolerant way, and for many programming tasks, you likely wouldn't have to worry about scaling and reconciling data in response to requests. It provides the benefits of the actor model through a programming model that attempts to abstract away details that you would otherwise have to worry about when using actors in production.
+
+# Why the actor model?
```
heathermiller, line 200: Don't worry about this section for now. We might even be able to omit it... Let's just get everything else taken care of and I'll figure out what can be done with this section afterwards.

chapter/3/message-passing.md
```diff
@@ -1,11 +1,250 @@
...
+
+The end result of all this decoupling and isolation of computation is that it is a natural fit for mobile ad hoc networks. In this domain, connections are volatile with limited range and transient failures. Removing coupling based on time or synchronization is a natural fit for the domain, and the communicating event-loop actor model is a natural model for programming these systems. AmbientTalk/2 provides additional features on top of the communicating event-loop model like service discovery. These enable ad hoc network creation as actors near each other can broadcast their existence and advertise common services that can be used for communication.
+
+AmbientTalk/2 is most notable as a reimagining of the communicating event-loops actor model for a modern use case. This again speaks to the broader advantages of actors and their applicability to solving the problems of distributed systems.
+
+# Active Objects
```
heathermiller, line 162: No mention of AmbientTalk/1 here? Fine by me if you have justification for excluding it. Just wondering what the rationale is.

chapter/3/message-passing.md
```diff
@@ -1,11 +1,250 @@
...
+
+I don't have an opinion on which of these is right. Both approaches have been successful in industry. Erlang has the famous use case of a telephone exchange and a successful history since then. Akka has an entire page detailing its usage in giant companies. Orleans has been used as a backend to massive Microsoft-scale games and applications with millions of users. It seems like the module approach is more popular, but there's only really one example of the managed runtime approach out there. There's no equivalent to Orleans on the JVM or Erlang VM, so realistically it doesn't have as much exposure in the distributed programming community.
+
+## Comparison to Communicating Sequential Processes (CSP)
```
heathermiller, line 240: Leaving it here is fine. The way you start this section is already a good way to connect it to everything else. That is, "you may be wondering why I didn't mention CSP. Well, that's because CSP is different, let me tell you why"

chapter/3/message-passing.md
```diff
@@ -1,11 +1,250 @@
...
+
+You might argue that I've ignored some other concurrency primitives that could be considered message-passing or actors at some level. After all, from a high level a Goroutine with channels feels a bit like an actor. As does an RPC system which can buffer sequential calls. I think a lot of discussions of actors are looking at them form a not-so-useful level of abstraction. A lot of the discussions of actors simply take them as something that is a lightweight concurrency primitive which passes messages. I think this view is zoomed out too far, and misses many of the subtleties that differentiate these programming models. Many of these differences stem from the flexibility and scalability of actors. Trying to use CSP-like channels to build a scalable system like you would an actor system would arguably be a tightly-coupled nightmare. The advantages of actors are around the looser coupling, variable topology, and focus on isolation of state and behavior. CSP has a place in building systems, and has proven to be a popular concurrency primitive, but lumping actors in with CSP misses the point of both. Actors are operating at a fundamentally different level of abstraction from CSP.
```
heathermiller, line 244: This section needs a bit more work. Perhaps give a basic description of CSP and point out how CSP can't be considered an actor framework.

chapter/3/message-passing.md
```diff
@@ -1,11 +1,250 @@
...
+
+# Modern usage in production
```
heathermiller, line 208: What are your plans for this section? Do you plan on citing instances in production systems where an actor system was chosen?

chapter/3/message-passing.md
```diff
@@ -1,11 +1,250 @@
...
+* _High performance_: Most industrial applications that use actors are highly performance sensitive, and high performance enables more graceful scalability.
+* _Flexible remote actors_: Many applications can benefit from remote actors, which can communicate transparently over the network. Flexibility in deployment mechanisms is also very important.
+
+These attributes give us a good basis for analyzing whether an actor system can be successful in production. These are attributes that are necessary, but not sufficient for an actor system to be useful in production.
+
+## Actors as a framework
+
+One trend that seems common among the actor systems we see in production is extensive environments and tooling. I would argue that Akka, Erlang, and Orleans are the primary actor systems that see real production use, and I think the reason for this is that they essentially act as frameworks where many of the common problems of actors are taken care of for you. This allows the programmer to focus on the problems within their domain, rather than the common problems of monitoring, deployment, and composition.
```
heathermiller, line 224: > many of the common problems of actors are taken care of for you

Such as? Be explicit.

heathermiller: > I would argue that Akka, Erlang, and Orleans

Just say:

> Akka, Erlang, and Orleans...

Remove the "I think"s, "I would say"s, etc as you pass through the article.

chapter/3/message-passing.md
```diff
@@ -1,11 +1,250 @@
...
+
+Akka and Erlang provide modules that you can piece together to build various pieces of functionality into your system. Akka provides a huge number of modules and extensions to configure and monitor a distributed system built using actors. They provide a number of utilities to meet common use-case and deployment scenarios, and these are thoroughly listed and documented. Additionally they provide support for Akka Extensions, which are a mechanism for adding your own features to Akka. These are powerful enough that some core features of Akka like Typed Actors or Serialization are implemented as Akka Extensions. Erlang provides the Open Telecom Platform (OTP), which is a framework comprised of a set of modules and standards designed to help build applications. OTP takes the generic patterns and components of Erlang, and provides them as libraries that enable code reuse and best practices when developing new systems. Cloud Haskell also provides something analogous to Erlang's OTP called the Cloud Haskell Platform.
```
heathermiller, line 226: > They provide a number of utilities to meet common use-case and deployment scenarios

Such as? Be explicit.

heathermiller: > Erlang provides the Open Telecom Platform (OTP)

New paragraph

chapter/3/message-passing.md
```diff
@@ -1,11 +1,250 @@
...
+
+Orleans is different from these as it is built from the ground up with a more declarative style and runtime. This does a lot of the work of distributing and scaling actors for you, but it is still definitely a framework which handles a lot of the common problems of distribution so that programmers can focus on building the logic of their system.
```
heathermiller, line 228: > This does a lot of the work of distributing and scaling actors for you, but it is still definitely a framework which handles a lot of the common problems of distribution so that programmers can focus on building the logic of their system.

Add a couple more sentences, be a little less vague about what Orleans handles for you

chapter/3/message-passing.md
```diff
@@ -1,11 +1,250 @@
...
+
+I don't have an opinion on which of these is right. Both approaches have been successful in industry. Erlang has the famous use case of a telephone exchange and a successful history since then. Akka has an entire page detailing its usage in giant companies. Orleans has been used as a backend to massive Microsoft-scale games and applications with millions of users. It seems like the module approach is more popular, but there's only really one example of the managed runtime approach out there. There's no equivalent to Orleans on the JVM or Erlang VM, so realistically it doesn't have as much exposure in the distributed programming community.
```
heathermiller, line 238: You can remove this sentences:

> I don't have an opinion on which of these is right.

heathermiller (commented):
Still very nice! :-) As you know from our last meeting, this is in great shape! I've left a ton of comments in the PR; most are suggested rewordings, so, most are small/quick fixes.

Otherwise, there are only a handful of biggish tasks left to do for the final revision. 

- Examples for each of the main kinds of actors. Some code showing how they are indeed different from one another is essential.
- Some diagrams for explaining some of the more difficult to understand descriptions (e.g., E)
- What happened to the failure-handling/monitoring section? Something ideally with a couple of examples discussing the different approaches. This stuff is immensely important for why real applications are being built using the actor model, so we should give a minimal understanding of this stuff in this chapter.
- Add citations inline throughout the paper (you have a bibliography, you just have to actually cite stuff now)
- Run through and remove all of the "I think"s, "I would argue"s, and "In my opinion"s throughout the paper. This should be pretty objective, since you've been writing based on your experience reading about each of these systems.

Anyway, other than the failure-handling/monitoring section that I thought we discussed in my office, the rest seems pretty manageable to do in the next few days. It's in great shape. Really nice work :-)

chapter/3/message-passing.md
```diff
@@ -1,11 +1,250 @@
...
-by: "Joe Schmoe and Mary Jane"
+title:  "Message Passing and the Actor Model"
+by: "Nathaniel Dempkowski"
 ---
 
-Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum. {% cite Uniqueness --file message-passing %}
+# Introduction
 
-## References
+Message passing programming models have essentially been discussed since the beginning of distributed computing and as a result message passing can be taken to mean a lot of things. If you look up a broad definition on Wikipedia, it includes things like RPC, CSP, and MPI. In practice when people talk about message passing today they mostly mean the actor model.
```
phaller, line 9: I would add a sentence or two that recognizes that the concurrency model of the pi-calculus is another important inspiration for practical message passing systems. For example, Go's channels are based on the idea of first-class communication channels from the pi-calculus. Then, Clojure's core.async library is based on CSP. You can then say that you focus on the actor model given its ubiquity and generality.

chapter/3/message-passing.md
```diff
@@ -1,11 +1,250 @@
...
 
-{% bibliography --file message-passing %}
\ No newline at end of file
+In the field of message passing programming models, it is not only important to consider recent state of the art research, but additionally the historic initial papers on message passing and the actor model that are the roots of the programming models described in newer papers. It is enlightening to see which aspects of the models have stuck around, and many of the newer papers reference and address deficiencies present in older papers. There have been plenty of programing languages designed around message passing, especially those focused on the actor model of programming and organizing units of computation.
```
phaller, line 11: "newer papers" -> "more recent papers"

chapter/3/message-passing.md
phaller: I'd suggest to cite Erlang and Akka as two industrial-strength actor systems.

chapter/3/message-passing.md
phaller: I would also add here that Gul Agha developed the model further, by focusing on using actors as a basis for concurrent (object-oriented) programming, adding a citation to Gul Agha's book (MIT Press).

chapter/3/message-passing.md
phaller: I would also mention that messages are sent asynchronously, which is actually the main reason why mailboxes are needed in the first place.

chapter/3/message-passing.md
phaller: I would even go so far as to say that the actor model is so general, because it places very few restrictions on systems. In particular, asynchrony and the absence of message delivery guarantees enables modeling real distributed systems using the actor model. (Requiring guaranteed message delivery would be much less general, it could only model distributed systems which include complex message-delivery protocols.)

chapter/3/message-passing.md
phaller: I agree with Heather about this being a bit confusing. I would state more explicitly whether you're talking about a stateful actor or not. Then you could explain what the state could be, in order to illustrate your point(s).

chapter/3/message-passing.md
phaller: I would be a bit more careful, and make sure to also look at, and refer to, Gul Agha's book (MIT Press), which was clearly the first step here, leading to his "Concurrent Object-Oriented Programming" paper.

chapter/3/message-passing.md
phaller: When you mention Rosette for the first time, make sure you add a citation to a good paper introducing Rosette. A good reference may be contained in the "Concurrent Object-Oriented Programming" paper. If not, you could simply cite "Concurrent Object-Oriented Programming" again, so that people know they can look there to find out more about Rosette.

chapter/3/message-passing.md
```diff
@@ -1,11 +1,250 @@
...
...d features, many of which we take for granted in  modern object-oriented programming languages. It implements dynamic creation and modification of objects for extensible and reconfigurable systems, supports inheritance, and has objects which can be organized into classes. I think the more interesting characteristic is that the concurrency in Rosette is inherent and declarative rather than explicit as with many modern object-oriented languages. In Rosette, the concurrency is an inherent property of the program structure and resource allocation. This is different from a language like Java, where all of the concurrency is very explicit. The motivation behind this declarative concurrency comes from the heterogeneous nature of distributed concurrent computers. Different computers and architectures have varying concurrency characteristics, and the authors argue that forcing the programmer to tailor their concurrency to the specific machine makes it difficult to re-map a program to another one. I think this idea of using actors as a more flexible and natural abstraction over concurrency and distribution of resources is an important one which is seen in some form within many actor systems.
```
phaller, line 65: Good sentence: "This is different from a language like Java, where all of the concurrency is very explicit." I suggest to add a citation to the book "Java Concurrency in Practice" (the current de facto reference).

chapter/3/message-passing.md
```diff
@@ -1,11 +1,250 @@
...
+* _Representation classes_ specify the resource management characteristics of implementations of abstract classes.
+* _Behavior classes_ specify the actual implementations of actors in given abstract and representation classes.
+
+These classes represent a concrete object-oriented abstraction to organize actors which handles the practical constraints of a distributed system. It represents a step in the direction of handling not just the information flow and behavior of the system, but the underlying hardware and resources. Rosette's model feels like a direct expression of those concerns which are something every actor system in production inevitably ends up addressing.
+
+## Akka
+
+Akka is an actively developed project built out of the work on [Scala Actors](#scala-actors) in Scala to provide the actor model of programming as a framework to Java and Scala. It is an effort to bring an industrial-strength actor model to the JVM runtime, which was not explicitly designed to support actors. There are a few notable changes from Scala Actors that make Akka worth mentioning, especially as it is being actively developed while Scala Actors is not.
```
phaller, line 77: "There are a few notable changes from Scala Actors": some important changes are detailed in the paper "On the Integration of the Actor Model in Mainstream Technologies-- The Scala Perspective" (http://dl.acm.org/citation.cfm?id=2414641)

chapter/3/message-passing.md
```diff
@@ -1,11 +1,250 @@
...
+
+Erlang actors run as lightweight isolated processes. They do not have visibility into one another, and pass around pure messages, which are immutable. These have no dangling pointers or data references between objects, and really enforce the idea of immutable separated data between actors unlike many of the early classic actor implementations in which references to actors and data can be passed around freely.
+
+Erlang implements a blocking `receive` operation as a means of processing messages from a processes' mailbox. They use value matching on message tuples as a means of describing the types of messages a given actor can accept.
+
+Erlang also seeks to build failure into the programming model, as one of the core assumptions of a distributed system is that things are going to fail. Erlang provides the ability for processes to monitor one another through two primitives:
```
phaller, line 108: "things are going to fail" -> "machines and network connections are going to fail"

chapter/3/message-passing.md
phaller, line 119: Just saying that I like your summary- well done!

chapter/3/message-passing.md
phaller: And here is the corresponding citation: "Akka.js: towards a portable actor runtime environment" (Gianluca Stivan, Andrea Peruffo, and Philipp Haller) :-) ACM DL: http://dl.acm.org/citation.cfm?id=2824821

chapter/3/message-passing.md
```diff
@@ -1,11 +1,250 @@
...
+
+When handing out references at a different level of granularity than actor-global, how do you ensure the benefits of isolation that the actor model provides? After all, by referencing objects inside of an actor from many places it sounds like we're just reinventing shared-memory problems. The answer is that E's reference-states define many of the isolation guarantees around computation that we expect from actors. Two different ways to reference objects are defined:
+
+* _Near reference_: This is a reference only possible between two objects in the same vat. These expose both synchronous immediate-calls and asynchronous eventual-sends.
```
phaller, line 137: Did you introduce the word "vat" before? I was surprised to read it here first, but perhaps I missed it earlier?

chapter/3/message-passing.md
```diff
@@ -1,11 +1,250 @@
...
+
+TODO: write more here, maybe something around promise pipelining and partial failure? implications of different types of communication? maybe mention some of the points that inspire some aspects of modern actors?
+
+## AmbientTalk/2
+
+AmbientTalk/2 is a modern revival of the communicating event-loops actor model as a distributed programming language with an emphasis on developing mobile peer-to-peer applications. This idea was originally realized in AmbientTalk/1 where actors were modelled as ABCL/1-like active objects, but AmbientTalk/2 models actors similarly to E's vats. The authors of AmbientTalk/2 felt limited by not allowing passive objects within an actor to be referenced by other actors, so they chose to go with the more fine-grained approach which allows for remote interactions between and movement of passive objects.
```
phaller, line 148: Not sure you mentioned ABCL/1 before, add a citation (Yonezawa et al).

phaller (commented):
A few comments. I hope they are useful. In general: great work, Nat!

chapter/3/message-passing.md
natdempk: That is a good idea, it would be nice to have a source to cite that also discussed the Java 8 world of concurrency which the book doesn't cover, but that's straying pretty far from message-passing and the core of this chapter.

chapter/3/message-passing.md
natdempk: Yes, it is introduced in the first paragraph. Might be worth playing with the formatting here similar to when the classic actor model is introduced so the term doesn't get glossed over.

heathermiller: I swear, if you get that picture in there, all confusion about what a "vat" is will go away. There are some nice ones in the  _Concurrency Among Strangers_ paper.

natdempk: True, I will do that.

chapter/3/message-passing.md
natdempk: So... I'm not sure what the context would be here exactly. I mention it briefly in the context of having more limitations on object references compared to AmbientTalk/2 using the communicating event-loops model in the AmbientTalk/2 section. I don't know if there are other points to be made independent of that which relate more to the idea of active objects, but I am open to ideas.

heathermiller: Yeah, perhaps there's nothing really compelling that can be added about AmbientTalk/1 other than "the first iteration of AmbientTalk was based on Active Objects". I guess you're right that it doesn't add much.

chapter/3/message-passing.md
natdempk: Added examples to hopefully clarify this section here: https://github.com/heathermiller/dist-prog-book/pull/12/commits/074a631ef668c6ffd08adb629862a6a2b5c2d3f2

chapter/3/message-passing.md
pgrosu: Maybe this paper might help [*Mirror-based Reflection in AmbientTalk*](http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.481.4964&rep=rep1&type=pdf) (below is a link):

http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.481.4964&rep=rep1&type=pdf

Here are some sections as comparison, that maybe might help:

*AmbientTalk/2 [10], an updated version of the language whose reflective API differs from the version presented in previous work [4] (pg. 5)*

*In previous work, we have discussed the metaobject protocol of AmbientTalk/1 – the predecessor of the AmbientTalk language described in this paper – to develop language features specifically for mobile ad hoc networks [4]. In AmbientTalk/1, an actor is represented as an active object which executes in a thread of its own, has a message queue and a dedicated behavior describing the methods that may be asynchronously invoked on the active object. This behavior object contains base-level application methods as well as metalevel methods used to hook into the actor’s metaobject protocol. Intercession is made possible by making the active object implement a metalevel method, which is only distinguishable from a base-level method by its reserved name. (pg. 34)*

*In AmbientTalk/1 reflection is neither stratified nor encapsulated: base-level code can be affected by the implementation details of metalevel code such as that of a new language feature. Because the base- and metalevels are not partitioned into separate namespaces, name clashes between the two levels can occur. For example, a base-level method may accidentally be regarded as a metalevel method simply because its name accidentally matches that of a metalevel operation. Finally, AmbientTalk/1 had no metaobject protocol for objects, only for actors. (pg. 34)*

chapter/3/message-passing.md
natdempk: I'm going to mention Akka, Erlang, and Orleans here. Are those the kind of examples you mean, or more along the lines of specific applications/systems build using those industrial actor frameworks?

heathermiller: Just given your sentence as it is,

> There are a few robust industrial-strength actor systems that are being used to power massive scalable distributed systems.

...I think you need both. You can keep it short, add a semicolon and make a list. Like,

> There are a few robust industrial-strength actor systems that are being used to power massive scalable distributed systems; for example, Orleans has been used to power the multiplayer experience of hit games like Halo (citation), Akka has been used to [...] (citation), and Erlang has been used to [...] (citation). 

Make sense?

natdempk: Yep, sounds good to me.

PR #7, opened 2016-12-10
Futures article

kisalaya:


heathermiller:
Merging so we can have it up on the site. Though I'll provide feedback in this (closed) PR :-) Just FYI

PR #12, opened 2016-12-15
Final working PR for Message Passing chapter

natdempk:
Quite a bit to do from https://github.com/heathermiller/dist-prog-book/pull/5 and in-person discussion. Opening a working PR to track progress.

Hit list, in addition to other PR comments:

- [x] different actor paradigm code examples
- [x] DIAGRAMS
- [x] more attention to monitoring aspect
- [x] Change Akka introduction to avoid history because its just confusing
- [x] Mention difference in process-based receive vs. previously mentioned Akka receive
- [x] In Scala actors section at the end mention that it has moved on to another life as Akka
- [x] More concrete examples of modules in the framework section. Bulleted list of the types/behaviors of modules
  * General examples of modules with short descriptions of what they do
- [x] What happened to the failure-handling/monitoring section? Something ideally with a couple of examples discussing the different approaches. This stuff is immensely important for why real applications are being built using the actor model, so we should give a minimal understanding of this stuff in this chapter.
- [x] Further explanation of monitoring and failure handling, what it really looks like to handle failure
  * Hello world of monitoring code snippet in some production thing
  * Make this more concrete
- [x] Try and back up high level generalizations with examples
- [x] Examples for each of the main kinds of actors. Some code showing how they are indeed different from one another is essential.
- [x] Some diagrams for explaining some of the more difficult to understand descriptions (e.g., E)
- [x] Add citations inline throughout the paper (you have a bibliography, you just have to actually cite stuff now)
- [x] Citation pass
  - Including non-published paper citations
- [x] Remove opinion sentence in managed runtime section
- [x] Run through and remove all of the "I think"s, "I would argue"s, and "In my opinion"s throughout the paper. This should be pretty objective, since you've been writing based on your experience reading about each of these systems.


natdempk:
This is it, calling it final at this point.

heathermiller:
Merging so it's live 😄 (I will PR my edits sometime in the next couple of days)

PR #17, opened 2016-12-17
Added images for counter example, need to work on text

aviralg:
Hi,

I got stuck in designing these diagrams for counters. A great deal of care and formatting has gone into this.
I don't have text as of now. This premature pull request is to take your feedback about the diagrams.
The diagrams show state and operation based specification of increment only and increment and decrement counters. 
For state based specification, I first show the false approach. Notice the spring arrows for state based approach which denote loss of message as state based approach works for lossy networks.
For both approaches, I have also shown concurrent operations.

The text is fairly straightforward to right now. I will refer to specific interesting points in the diagram and talk about what's happening there. 

The specification of the counters will be pythonesque in nature.

Let me know if you have issues with font size, color or maybe the execution of operations, labeling of parts, arrows and other formatting related issues.


heathermiller:
Wow, these diagrams are amazing! They're beyond anything I would have expected. I have no feedback other than "they're great!". Your approach seems totally sound to me.

I'm a bit concerned about the state of the text though; it seems you've got a lot to do to explain what's going on in the diagram in text. What's your plan?

aviralg:
Thanks. The explanation part is easier now. The next commit will have text and specification for all of these diagrams. Having a diagram makes it really easy to write the text. Certain things can only be explained visually. Points such as - why commutativity is important, or messages can be lost etc. can be easily explained by referring to the diagram.

heathermiller:
Sure, this is all true, and I agree with you. Following the explanation of the diagrams, you'll need to tie everything up and conclude (the three sub-chapters) though. That'll take some thought as well, of course.

pgrosu:
Hey Aviral,

I concur with Heather, the diagrams are awesome, and illustrate the ideas well.  I agree that it'll be straightforward now to tie the fundamentals together through them to make it a nice story - you're really good at that!  I remember the clarity with which you explained the stuff a month ago, when covering the second Shapiro paper, so it should be no problem for you :)

Cool man - look forward to reading it!
~p

aviralg:
Added specification of counters. Working on description.

aviralg:
**Please read from PN Counter onwards**.  Don't even read the material before that. Its just a mess. Once PN Counter is finalized, the same description (with minor edits) will be copied to increment only counters. Not happy with what I have written, does not look professional enough. But its better to take your feedback at this early stage.

pgrosu:
@aviralg It's starting to come together nicely!   So just looking at the PN-counter portion, it might be best to start with why you are introducing it, and how it connects to the previous counters - basically it has to flow.  Then I would describe it in general terms with a helpful figure, and then augment it using the helpful example.  For some reason the figure is not showing below it to put the dots together - might be a formatting thing, but probably will show up in the formatted html when it's merged into master.  

This is just my initial $0.02 recommendation, but @heathermiller is the boss :)

Hope it helps,
~p

heathermiller:
@aviralg, it looks good! Keep working at it; I like the approach you're taking, it's **very** intuitive! It's not often that people invest as much effort as you are investing in trying to make intuitive sense of different sorts of CRDTs. This is shaping up to be a pretty meaningful contribution if you ask me! 

The writing is OK. It's a bit short in some places, but we can work on it. Just keep going.

heathermiller:
PS, I agree with @pgrosu about needing to better connect sections, but I assume this is missing only because you asked to start from the PN-counter section onward, so I was assuming you'd get to this stuff later.

aviralg:
Completed the writeup. Let me know your views.

pgrosu:
borat_great_success

aviralg:
A man't gotta do what a man's gotta do. Took me a while to do this, more from the desire of solving a puzzle than anything else. But its important to show how the decrement operation takes the state up in the lattice. Still not happy with the topology. Wondering if I can represent it as stacked cubes or some other regular geometrical figure. But I enjoyed making this. I recommend you try this too.

cmeiklejohn:
Nice images!

janisz (approved):

PR #23, opened 2017-01-02
Replace communication with communicate

tiagocoutinho:
Small typo: _communication information_ should be _communicate information_


heathermiller:
Thank you!

PR #25, opened 2017-01-04
Fixes to futures chapter

heathermiller:
cc/ @kisalaya89 

Also, @kisalaya89, you didn't cite your references in-line. That is, citations for some block of text should look like:

> Blah de blah [(Smith 2012)]() is a tool that blah blah, and more blah blah. 

Or just check another chapter to see what I mean. E.g, http://dist-prog-book.com/chapter/3/message-passing.html

There are examples of how to do this in example.md. When you mention a language or tool or make a point that has something to do with one of your references, you cite it; you just write something like: `{% cite joeJoseph14 %}`, and then the link + reference gets automatically generated for you.

I would go through and do this for your writeup on my own, but there are some places where you make a point and I can't know specifically which reference you meant to cite. Obvious stuff I can help with, like citing E or some language or website that you explicitly mention. But there are some cases where I can't connect a point to a reference and I might need your help here. (Shouldn't take long)


heathermiller:
PS, more fixes to come. Don't review yet.

heathermiller:
The "Motivation" section is really more of a "History" section. Thus I'm changing around the order of some of these paragraphs because I think it would make more sense if the definitions section came after the motivation, but before the history.

heathermiller:
Ok, I think those are all the big changes I planned on making. I shuffled things around a bit and added a basics section, material on E, and more. 

So I did a lot of massaging the first half up until the end of the history section. I've also fixed a lot of bibliography stuff. 

Though I admit I still don't know what to do with many of the bibtex entries in the futures bibtex file... I don't know if there are blog articles you had meant to cite when you made some point in the text that you had read in a blog post. Help fixing the bibliography would be most welcome.

In any case, @kisalaya89 could you have a quick look through what I've added to let me know if you have any major objections?

heathermiller:
(I'm merging this so it's easier to read/review... Of course you're still welcome to comment here on the PR whether it's open or closed.)

heathermiller:
@kisalaya89, here's a link to the page for easier reading: http://dist-prog-book.com/chapter/2/futures.html

PR #41, opened 2017-10-25
Proposed revision to 4-5

semaj:
@natdempk @heathermiller just a rough outline for the merging and revision of these two chapters. feedback plz. I suggest just viewing the file in rendered MD.

### DIFF
--- a/chapter/4/dist-langs.md
+++ b/chapter/4/dist-langs.md
@@ -4,513 +4,110 @@ title:  "Distributed Programming Languages"
 by: "Connor Zanin"
 ---
 
-## Problems of Distributed Programming
+# Why is Distributed Computing so hard?
 
 There are problems that exist in distributed system environments that do not exist in single-machine environments.
-Partial failure, concurrency, and latency are three problems that make distributed computing fundamentally different from local computing.
-In order to understand the design decisions behind programming languages and systems for distributed computing, it is necessary to discuss these three problems that make distributed computing unique.
-In this section, we present an overview of these three problems and their impact on distributed programming models.
 
-### Partial Failure
+## Node Failures
 
-In the case of a crash on a local environment, either the machine has failed (total failure), or the source of the crash can be learned from a central resource manager such as the operating system. {% cite waldo1997 --file dist-langs.bib %}
-If an application consists of multiple communicating processes partial failure is possible, however because the cause of the partial failure can be determined, this kind of partial failure can be repaired given the operating system's knowledge.
-For example, a process can be restored based on a checkpoint, another process in the application can query the operating system about the failed process' state, etc.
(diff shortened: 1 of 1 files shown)


natdempk:
So I think we simply have too many ideas here for a single chapter.

Maybe we could split them along things that apply to every chapter vs. things that apply to specifically distributed languages + extended languages?

This is hard, as it begs he question of how to actually make that split. To me it feels like there are things that are "foundational problems of distributed computing" and there are things that are closer to the design-space decisions that lead to solving the problems within the domain of distribution.

Ideas that fall under the spectrum of foundational problems:

- Distribution as a computing problem domain
- Node Failures
- Packet Drops / Delays / Corruption
- Network Partitions
- Latency
- The CAP Theorem

Obviously this stuff needs to be mentioned in a book about programming languages trying to solve these problems, but in my opinion these almost feel like their own starting chapter or whatever to setup the book. These problems drive all distributed programming models and language design concerns, so to me they feel like something larger that should be addressed independently.

The next set of ideas are around how to tackle these problems in practice, and make it so that people can flexibly (or not) build systems that solve these problems. This gets us into the space of ideas like:

- Build vs. extend a language? Why take one vs. the other approach?
- Approaches to building/extending languages
  - Distributed shared memory
  - Virualization models (actor/object/method spectrum)
  - Dataflow
- Implementation level + associated tradeoffs
  - OS
  - Compiler
  - DSL/Library
- Design/abstraction level comparisons
  - Abstraction over shared data
  - Abstraction over objects/methods
  - Abstraction over general message-based computation

For each of these ideas, they obviously tie by in to solving the core problems of distributed computation. Each of these solves or doesn't solve different problems around things like:
- fault tolerance
- state locality

This is the oportunity to bring up all the languages that have explored these different approaches and feel out the important details of each one.

I think some of these make sense to highlight different trends as time goes on, leading to the "modern" approaches to tacking the problems of distribution. This could discuss the few models that seem to have won out at a language-level which are effictively DSLs and libraries, with some of these approches like futures etc. getting pushed down in the languages themselves. 

I guess at the end of this my proposal would be two chapters:

1. Defining the domain of distribution: why do we need programming models specifically for distributed computing?
2. The distributed language spectrum: what's the design space that distributed programming "languages" have explored?

The two sets of bullets above could be rough outlines of the core ideas of the two chapters. I think the first one could be a lighter chapter, and the second could be this heavier exploration of the design space of languages.

semaj:
I suppose my proposed outline could be split in half to produce the two chapters you suggest. I worry that the second chapter will be much longer than the first. This may be okay, but the domain of distribution is more of an intro whereas there's *so* much to talk about when it comes to proposed distributed languages and their tradeoffs.

In any case, I'd love to write the domain of distribution chapter.

